PR #10, opened 2025-12-11
 feat: add basic audio support with voice recording and TTS

PaulLampe:
## Summary

- Add voice recording to send audio messages that get transcribed
- Add text-to-speech (TTS) playback for assistant messages
- Integrate audio controls into chat input and message components

## Additional changes

- Applied formatting to all existing files 
- Added pre-commit hook via Husky for formatting and linting
- Minor import organization fixes


Antonwy:
> - [ ]  Merge into develop instead of main

I think this PR should be based on #11 so the diff is smaller. after #11 is merged, the base will automatically switch to `main`.

PR #30, opened 2026-01-11
add: updated how-to content and pdf export. 

RobinFrasch:
- content is now in variables to ensure consistency between page and exported pdf
- updated settings.json because of linter headache


components/how-to.tsx
```diff
@@ -18,324 +22,604 @@ import {
...
-    'Wer steht hinter wahl.chat?',
-    'Wie funktioniert Briefwahl?',
-  ];
+  const [isExporting, setIsExporting] = useState(false);
 
   const buildQuestionLink = (question: string) => {
     return `/session?q=${question}`;
   };
 
+  const exportToPDF = async () => {
```
romman8: Could you move the PDF export logic to a separate file to improve maintainability?

PR #48, opened 2026-01-30
remove: data collection on wahl-agent

SimonKaran13:
### TL;DR

Removed user data collection functionality from the agent chat flow.

### What changed?

- Removed the `DataCollectionForm` component and its references
- Removed the `data-collection` step from the agent flow
- Updated the `createConversation` function to no longer require user data
- Removed user data related state and actions from the agent store
- Removed the data collection section from the consent screen
- Modified the flow to go directly from conversation choice to topic selection

### Why make this change?

This change simplifies the user experience by removing the personal data collection step. The data collection will happen in the study's questionnaire on Uni Park


RobinFrasch (approved):
lg, no changes to functionality

PR #105, opened 2026-03-11
Update main with latest changes

romman8:


SimonKaran13:
We did the first round of the study for my thesis in the dev deployment with 30 participants and everything worked fine. So this time it is actually quite well tested xD

PR #111, opened 2026-07-14
wahl.chat V2:  Data Layer & Foundation

annikasiefke:
# wahl.chat V2: Data Layer & Foundation

This MR replaces the per-election content silos of V1 with a single, continuously updatable corpus and the pipeline to feed it and migrates some of the platform's foundations (transport, HTTP framework, Python packaging). The chat experience itself stays functionally the same as V1, but answers are now grounded in live parliamentary data and cited sources open in-page as videos and PDFs instead of bare links.

## The core idea: one corpus instead of one collection per election

V1 duplicated content into a Qdrant collection per election context, so shared national documents existed many times and cross-election queries needed multiple round-trips. V2 stores everything in **one Qdrant collection per environment** (`wahlchat_chunks_<env>`) and scopes queries with payload filters instead:

- **`party_id` as tenant key** (Qdrant `is_tenant` indexing): vectors are co-located per party; every query filters by party, so no global HNSW graph is built at all.
- **`region` filtering**: each chunk carries the region it belongs to; at query time the election context supplies its region path (e.g. federal → state → district) and matching is a single `MatchAny` filter. State-election chats automatically see national documents and vice versa where appropriate.
- **`authority_tier`**: every chunk is ranked `authoritative` / `factual_record`/ `self_reported` / `promotional`, so retrieval can prefer official records over campaign material. Currently each source's tier is only surfaced to the LLM in the prompt ("Vertrauensstufe") so it can weigh source trust when answering.
- **Idempotent ingestion**: deterministic point IDs plus a `content_hash` per chunk mean re-running a connector is always safe: unchanged content is skipped cheaply, changed content is rewritten and orphaned chunks from shrunken documents are cleaned up.

The shared data contract lives in `src/ingestion/schemas.py` (`ChunkRecord`, frozen Pydantic, `extra="forbid"`) -> common ground that lets the team build further connectors in parallel without diverging.

## The ingestion pipeline

Every connector implements the same three-method interface (`discover → fetch → normalize`) and is pure data transformation. A shared runner (`src/ingestion/run.py`) owns everything else: 
- incremental cursors (derived from what's already in Qdrant, so there is no separate watermark store to corrupt)
- embedding (OpenAI `text-embedding-3-large`, unchanged from V1 for index parity) -> this soon will change to Google Enterprise AI
- batched upserts
- per-item skip-and-warn error handling. 
Batch windows are sized so runs fit scheduler limits; connectors should be deployed as Cloud Run Jobs since long backfills exceed Firebase's 15-minute scheduled-job cap.

### The connectors

| Connector                    | Source                                    | What it ingests                                                                                                                                                                                                                                                                               |
| ---------------------------- | ----------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Abgeordnetenwatch votes**  | abgeordnetenwatch.de API                  | Roll-call votes (Bundestag + Landtage) as per-party tallies (`vote_record`). Incremental via poll-ID cursor.                                                                                                                                                                                  |
| **Bundestag speeches (DIP)** | Bundestag DIP API                         | Plenary speeches parsed from protocol XML (`parliamentary_speech`). Speaker→party resolution via the official MdB master data. Date-cursor incremental, designed for daily/weekly refresh runs.                                                                                               |
| **openparliament.tv**        | openparliament.tv bulk data               | The same plenary speeches, but with sentence-level timing and a **video deep-link** per speech. Coexists with DIP: duplicates are detected via a shared speech key, and the richer op.tv version supersedes its DIP twin (grafting the DIP transcript PDF onto it first, so nothing is lost). |
| **Manifestos**               | abgeordnetenwatch.de election-program API | Party manifestos for all federal and state elections since 2020, scraped from the linked HTML or parsed from the PDF. Citations always point at the original publisher — we never re-host files.                                                                                              |


## Chat: cited sources become viewable media

Clicking a source now opens an in-page viewer dialog:

- **Speech citations open the actual plenary video**, seeked to the cited moment. The backend refines each op.tv speech citation to a sentence-level timestamp (picking the most relevant cited sentence) and the frontend plays the video from exactly there.
- **PDF citations (manifestos, transcripts) open an in-page PDF viewer** deep-linked to the cited page (if present). A search-term fragment for the cited passage is also passed, but highlighting it is browser-dependent and unreliable in practice — the page deep-link is the dependable part. PDFs are served through a same-origin proxy with a host allowlist; touch devices fall back to a new tab.
- The **sources dialog** ("Quellen"), **voting-behavior detail view** (per-party roll-call results with source links) and **pro/con perspectives** carry over from V1, with voting behavior now backed by the real ingested `vote_record` corpus instead of pre-baked data.

## Platform migrations (and why)

- **Socket.IO → SSE** (FastAPI + `sse-starlette`, Vercel AI SDK v5 `useChat` on the frontend). we discussed that in one of the initial meetings and also Websockets are blocked by some corporate networks, while SSE is plain HTTP. The aiohttp app is replaced by FastAPI (typed routes, Pydantic request validation, OpenAPI docs). 
- **Poetry → uv** for the Python backend: dramatically faster installs, standard PEP 621 metadata, reproducible `uv sync --locked` in CI.
- **Fully local dev stack**: `make stores-up` starts Qdrant (docker-compose) and the Firebase emulators, `make seed-local` seeds Firestore, `make dev-local` runs web + backend against local stores only — nothing during development touches real data.

## Compliance

GDPR Art. 9 (political opinions are special-category data) is enforced structurally: no ingestion code path reads user data (guarded by a static test) and the party-answer cache only stores answers from curated conversations (proposed questions and their quick-reply follow-ups), free-form user conversations are never cached.

## Known limitations / deferred decisions

- Multi-party streaming is serialized.
- Voice/audio chat input was dropped together with the websocket transport; the `/api/v1/tts` endpoint is currently non-functional (no web caller exists for either).
- Vote justifications: V1 showed AI-generated justification texts produced offline; V2 grounds voting behavior in real roll-call tallies and does not generate justifications, so the justification panel currently never renders. Should we regenerate them as clearly-labeled AI enrichment or just remove the UI? 
  **An idea for this:** rebuild them grounded instead of free-floating, link each vote to its plenary debate (AW poll <-> DIP Drucksache/`fundstelle`), retrieve the party's speeches from that debate and have the LLM summarize the party's *stated* reasons as a clearly-labeled AI summary with citations deep-linking into the speech videos.
- The parliamentary-questions endpoint is dormant (no corpus data yet).
- Bundestag-speech transcript PDFs (the full plenary protocols) currently open without a page reference: we ingest the speeches from the protocol XML, not the PDF, so no per-speech page number is stored and the search-fragment fallback is unreliable. A fix we could try is to derive the page from the DIP `fundstelle` page numbers at ingest time and deep-link `#page=` like manifestos.

## Testing

Backend: 523 tests passing (includes regression tests for cursor/watermark behavior, orphan cleanup, and the cache gate), `ruff` and `mypy` clean on `src/`. 
Web: `tsc --noEmit` clean, production `next build` succeeds. 
Retrieval verified live against a local corpus of ~30k chunks (votes, speeches, manifestos).


romman8:
> Vote justifications: V1 showed AI-generated justification texts produced offline; V2 grounds voting behavior in real roll-call tallies and does not generate justifications, so the justification panel currently never renders. Should we regenerate them as clearly-labeled AI enrichment or just remove the UI?
An idea for this: rebuild them grounded instead of free-floating, link each vote to its plenary debate (AW poll <-> DIP Drucksache/fundstelle), retrieve the party's speeches from that debate and have the LLM summarize the party's stated reasons as a clearly-labeled AI summary with citations deep-linking into the speech videos.

Let's remove them in this MR and bring them back as clearly labeled AI summaries with links to the transcript / video source in a follow-up MR. Wdyt?

README.md
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
 
 # wahl.chat
 
-The leading political information chatbot for the German federal elections 2025.
+A grounded, source-cited political-information chatbot for German elections — federal, state, and local.
```
romman8: nit: 
1. should be "political information" instead of "political-information"
2. only "grounded" is enough imo --> can remove "source-cited"

CLAUDE.md
romman8, line 1: This CLAUDE.md documents the agent instructions for the transition from the initial version of wahl.chat.
However, before we merge this, please change it into a CLAUDE.md that rather documents the architecture of the system after this MR and how to work with it (how to run it, how to treat data, coding guidelines, and how to test/verify changes, etc.)

romman8: Also another small thing that would be great:
Let's use `AGENTS.md` instead of `CLAUDE.md` and simply reference `AGENTS.md` from `CLAUDE.md` or (my preferred approach) create a symlink from provider specific `CLAUDE.md` to `AGENTS.md` in a small setup script (https://www.reddit.com/r/ClaudeAI/comments/1na5dy8/use_symlinks_to_connect_claudemd_agentsmd/)

.gitignore
```diff
@@ -1,18 +1,33 @@
+# === Planning (local only) ===
+.planning/
+# Agent-local runtime state (worktrees, locks). NOTE: CLAUDE.md itself is
+# committed and must NOT be ignored.
+.claude/
```
romman8: we should not ignore the whole `.claude/` directory. It can also hold skill that should be committed to the repo

.gitignore
```diff
@@ -1,18 +1,33 @@
...
 .env.*
 !.env.example
 *firebase-adminsdk*.json
 
+# === Python runtime ===
+__pycache__/
+*.py[cod]
+*.pyo
+.venv/
+*.egg-info/
```
romman8, line 36: Why is this part of the top-level gitignore rather than the backend-specific gitignore?

.pre-commit-config.yaml
romman8, line 1: Why does this remove pre-commits?
Let's definitely keep them

.github/workflows/ci.yml
```diff
@@ -0,0 +1,250 @@
+# CI pipeline — wahl.chat V2 (D-08, D-09, FOUND-05)
+#
+# Jobs:
+#   backend        — GDPR wall guard + uv ruff check + mypy + pytest (unit/integration, not smoke)
```
romman8: nit: let's split backend linting and testing into separate jobs

.github/workflows/ci.yml
```diff
@@ -0,0 +1,250 @@
...
+#   frontend       — bun frozen install + bun audit (D-08 supply-chain gate) + lint + tsc + build
+#   firebase-rules — Firestore security-rules tests against the emulator (G2)
+#   smoke-test     — D-09 E2E SSE smoke test (needs: [backend, frontend])
+#
+# Key constraints (CLAUDE.md):
+#   - uv sync --locked  everywhere (locked installs only — no version bumping)
+#   - bun install --frozen-lockfile  (never bare bun install in CI)
+#   - bun audit gate (D-08 supply-chain hardening)
+#
+# Smoke-test determinism (D-09 / plan 01-08 Task 1):
```
romman8: What are these `D-09 / plan 01-08 Task 1` references?
Let's clean them up if they reference plans/specs that are not part of the codebase

.github/workflows/ci.yml
```diff
@@ -0,0 +1,250 @@
...
+
+name: CI
+
+on:
+  push:
+  pull_request:
+
+# A push to a branch that already has an open PR fires BOTH the `push` and the
+# `pull_request` events, double-running the pipeline. The `guard` job below lets
+# exactly one proceed: the push run when there is NO open PR for the branch, and
```
romman8: What is the motivation for running on every push?
Imo only on PRs suffices

.github/workflows/ci.yml
```diff
@@ -0,0 +1,250 @@
...
+
+      # Reproducible install — never bare bun install in CI (CLAUDE.md)
+      - name: Install dependencies (frozen)
+        run: bun install --frozen-lockfile
+
+      # D-08: supply-chain hardening gate (CLAUDE.md "bun audit as CI gate")
+      # HARD-BLOCKING at --audit-level high: there are currently 0 critical/high
+      # findings (Next.js bumped to 15.x >=15.2.3, fixing GHSA-f82v-jwr5-mffw;
+      # unused `swiper` removed, fixing the GHSA-hmx5-qpq5-p643 prototype-pollution
+      # critical). Any new high/critical fails CI.
```
romman8: prefer timeless comments over specific version upgrades that were made in this PR. These changes are documented via commit history

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/stance.py
```diff
@@ -0,0 +1,195 @@
...
+# SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
+
+"""
+Pure stance-derivation and fraction-aggregation mappers for the AW connector.
+
+All functions in this module are I/O-free (no network, no Firestore, no
+filesystem) and deterministic — suitable for unit testing without mocks.
+
+build_qsp() and QuestionStancePair removed — src/matcher/
+deleted; these functions produced Firestore matcher objects that the
```
romman8: nit: avoid these comments referencing old variants of the code produced during your development process

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/stance.py
```diff
@@ -0,0 +1,195 @@
...
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+_PARTY_SLUG_QUARANTINE = "unbekannt"
```
romman8, line 37: nit: let's use English in the code for anything that is not specific to the German elections. I.e. "unknown" here

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/stance.py
```diff
@@ -0,0 +1,195 @@
...
+# than crashing the run (skip-and-continue, do NOT crash).
+_VALID_VOTE_TOKENS = ("yes", "no", "abstain", "no_show")
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def aggregate_fraction_tallies(votes: list[dict]) -> dict[int, dict]:
```
romman8: nit: would prefer a more strongly typed return value (e.g. dict[int, FractionTally])

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/stance.py
```diff
@@ -0,0 +1,195 @@
...
+       yes -> "agree", no -> "disagree", abstain -> "neutral", no_show -> "absent".
+
+    Args:
+        yes: Count of "yes" votes in the fraction.
+        no: Count of "no" votes.
+        abstain: Count of "abstain" votes.
+        no_show: Count of "no_show" votes.
+
+    Returns:
+        One of: "agree", "disagree", "neutral", "absent".
```
romman8: nit: let's introduce an enum type for the stance to make the possible values explicit

ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
```diff
@@ -0,0 +1,415 @@
...
+        legislature_id: Optional[int] = None,
+    ) -> None:
+        # Env read lives in the __init__ BODY (D7): an import-time default arg
+        # would freeze the value for the process lifetime and force one process
+        # invocation per legislature.
+        if legislature_id is None:
+            legislature_id = int(os.getenv("AW_LEGISLATURE_ID", "111"))
+        self._legislature_id = legislature_id
+
+        # Validate legislature_id against LEGISLATURE_CONFIG; fail loudly on miss.
```
romman8: nit: let's avoid comments like this one that purely describe what the code does without any rationale (the other comments here are mostly fine as they explain the semantics of a variable or provide rationale

ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
```diff
@@ -0,0 +1,415 @@
...
+                    key="source_type",
+                    match=qdrant_models.MatchValue(value="vote_record"),
+                ),
+                qdrant_models.FieldCondition(
+                    key="legislature_period_id",
+                    match=qdrant_models.MatchValue(value=self._period_id),
+                ),
+            ]
+        )
+        while True:
```
romman8, line 182: What is the order of magnitude of polls that could be retrieved here?

ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
```diff
@@ -0,0 +1,415 @@
...
+                f"foreign-legislature poll (skipping)."
+            )
+
+        poll_date_str: str = poll.get("field_poll_date") or ""
+        try:
+            publish_date = date_type.fromisoformat(poll_date_str)
+        except (ValueError, TypeError):
+            # skip-and-warn instead of fabricating date.today().
+            # A missing/unparseable poll date makes the publish_date non-deterministic;
+            # raising ValueError here triggers run_connector skip-and-continue.
```
romman8: nit: these comments about raising to trigger skip-and-continue are super repetitive and are bloating the codebase for both humans and agents + creating a bad example of code comment style for agents --> would prefer to remove these

ai-backend/src/ingestion/connectors/abgeordnetenwatch/legislature_config.py
```diff
@@ -0,0 +1,255 @@
...
+#
+# To add a prior legislature period: add a row here. No other code change needed.
+# To add a new Landtag election: add the new parliament_period row from the AW API
+# (GET /parliament-periods?parliament={parl_id}&type=legislature) and update
+# the existing row's date_to to the end_date_period of the outgoing period.
+# ---------------------------------------------------------------------------
+LEGISLATURE_CONFIG: dict[int, LegislatureConfig] = {
+    # -----------------------------------------------------------------------
+    # Bundestag — wahlperiode is the sole source of the Wahlperiode number
+    # (connector.py's former _LEGISLATURE_TO_WAHLPERIODE map is deleted).
```
romman8: nit: don't reference dead code

ai-backend/src/ingestion/connectors/abgeordnetenwatch/legislature_config.py
```diff
@@ -0,0 +1,255 @@
...
+      1. If ``legislature_period_id`` names a row in :data:`LEGISLATURE_CONFIG`,
+         use that row — unambiguous. Covers federal contexts and any context
+         that already names its term (AW vote-record precision filter).
+      2. Otherwise, if ``region_path`` is set, take the most-specific element
+         (``region_path[-1]``) and gather rows whose ``region`` equals it:
+           - none  → return ``None`` (caller falls back to single-pass).
+           - one   → use that row.
+           - many  → deterministic tiebreak (see below).
+
+    Multi-period tiebreak (e.g. DE-BW / DE-RP after the D6 outgoing rows were
```
romman8: What does D6 reference?

ai-backend/src/ingestion/connectors/bundestag_speeches/mappers/corpus.py
```diff
@@ -0,0 +1,378 @@
...
+
+    Target ≤ max_tokens tokens per chunk (well under the 8191-token
+    OpenAI embedding limit). Short texts that fit in a single chunk
+    are returned as-is (list of one string).
+
+    Args:
+        text:       Full speech text to chunk.
+        max_tokens: Maximum token count per chunk (default 6000).
+        overlap:    Number of tokens to repeat at the start of the next
+                    chunk for context continuity (default 200).
```
romman8: How did you chose this chunking strategy and its parameters?
1. Why token-based over a character-based strategy?
2. The chunk size of 6000 tokens seems very large. How did you chose that?
3. The overlap of 200 seems very small compared to the chunk size. Usually an overlap of 10-20% is recommended (see e.g. here: https://www.firecrawl.dev/blog/best-chunking-strategies-rag)

romman8: note: feel free to use pre-existing chunking functions from the langchain library here

ai-backend/src/ingestion/connectors/bundestag_speeches/bulk.py
```diff
@@ -0,0 +1,236 @@
...
+
+if __name__ == "__main__":
+    # Load ai-backend/.env if present.
+    # bulk.py sits at: ai-backend/src/ingestion/connectors/bundestag_speeches/bulk.py
+    # parents[4] = ai-backend/ (same depth as manifestos/bulk.py)
+    from dotenv import load_dotenv  # noqa: PLC0415
+
+    _env_path = Path(__file__).resolve().parents[4] / ".env"
+    if _env_path.exists():
+        load_dotenv(_env_path, override=False)
```
romman8: is override=False intentional? I usually prefer override=True to have the .env as the source of truth for env variable values

ai-backend/src/ingestion/connectors/bundestag_speeches/client.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+
+    def __init__(self, api_key: str, sleep: float = 0.2) -> None:
+        self.api_key = api_key
+        self.sleep = sleep
+        self.session = requests.Session()
+        self.session.headers.update(
+            {
+                "Accept": "application/json",
+                "User-Agent": "Mozilla/5.0 bundestag-speeches-student-project/1.0",
```
romman8: how did you chose this user agent? We can also use the wahl.chat user agent here that you used for Abgeordnetenwatch

ai-backend/src/ingestion/connectors/bundestag_speeches/client.py
```diff
@@ -0,0 +1,184 @@
...
+            if ".enodia/challenge" not in response.url:
+                break
+            if attempt == 2:
+                data = curl_get(url, accept="application/json")
+                time.sleep(self.sleep)
+                return json.loads(data)
+            time.sleep(5 * (attempt + 1))
+
+        response.raise_for_status()
+        time.sleep(self.sleep)
```
romman8, line 113: Why do we sleep within the get if there is no limit on sequential requests?

ai-backend/src/ingestion/connectors/bundestag_speeches/client.py
```diff
@@ -0,0 +1,184 @@
...
+        Raises:
+            requests.HTTPError: on non-2xx responses (after all attempts).
+            DipChallengeError: if curl fallback also hits the challenge page.
+        """
+        params = dict(params or {})
+        params["apikey"] = self.api_key
+        params["format"] = "json"
+        url = BASE_URL + endpoint + "?" + urlencode(params, doseq=True)
+
+        for attempt in range(3):
```
romman8, line 102: Can we also use `Retry` from `urllib3.util.retry` for retry logic like you did for Abgeordnetenwatch and just add special handling for the ".enodia/challenge" case?

ai-backend/src/ingestion/connectors/bundestag_speeches/constants.py
```diff
@@ -0,0 +1,69 @@
...
+
+# Hard page cap for DipClient.pages() (mirrors the AW client's _MAX_PAGES).
+# Rationale: a full 20th Bundestag legislature (~4 years) has ~130 plenary protocols,
+# each yielding roughly one DIP API page. 500 pages is more than enough for any
+# plenary-protocol listing while preventing an infinite loop on a stuck/malformed cursor.
+_MAX_PAGES = 500
+
+# MdB-Stammdaten XML path (relative to ai-backend/).
+# Download: https://www.bundestag.de/resource/blob/472878/MdB-Stammdaten.xml
+#           → ai-backend/data/MDB_STAMMDATEN.XML  (gitignored, 15 MB)
```
romman8: Would be great if the download to the correct dir was part of some setup script

ai-backend/src/ingestion/connectors/bundestag_speeches/mdb.py
```diff
@@ -0,0 +1,126 @@
+# SPDX-FileCopyrightText: 2025 wahl.chat
+#
+# SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
+
+"""MdB-Stammdaten lookup — load speaker→party mapping from the Bundestag master file.
+
+Absorbed from ~/Downloads/bundestag_speeches/mdb.py.
```
romman8: What does this mean?

ai-backend/src/ingestion/connectors/bundestag_speeches/parser.py
```diff
@@ -0,0 +1,169 @@
...
+                continue
+
+            text = paragraph_text(child)
+            if text:
+                paragraphs.append(text)
+
+        rede_id = speech_node.get("id")
+        speeches.append(
+            {
+                "xml_rede_id": rede_id,
```
romman8, line 171: nit: let's keep variable names to English. I.e. `xml_speech_id`

ai-backend/src/ingestion/connectors/manifestos/connector.py
```diff
@@ -0,0 +1,452 @@
...
+        date_str: Optional[str] = data.get("election_date") or data.get("start_date_period") or None
+        cache[period_id] = date_str
+        return date_str
+    except Exception as exc:  # noqa: BLE001
+        logger.warning("Could not resolve parliament-period %s: %s", period_id, exc)
+        cache[period_id] = None
+        return None
+
+
+def determine_source(program: dict) -> tuple[str, str]:
```
romman8: nit: let's introduce an enum for the source kind

ai-backend/src/ingestion/connectors/manifestos/connector.py
```diff
@@ -0,0 +1,452 @@
...
+            resp.raise_for_status()
+            text_content = extract_main_text(resp.text)
+        except Exception as exc:  # noqa: BLE001
+            raise ValueError(f"HTML fetch failed: {exc}") from exc
+        if len(text_content) < 500:
+            raise ValueError(f"HTML too short ({len(text_content)} chars)")
+        return {"pages": [(1, text_content)], "total_pages": None}
+
+    # source_kind == "pdf"
+    from io import BytesIO  # noqa: PLC0415
```
romman8: why is this import only local within the function?

ai-backend/src/ingestion/connectors/manifestos/mappers/corpus.py
```diff
@@ -0,0 +1,540 @@
...
+# Party slug map — for election-program party.label
+# ---------------------------------------------------------------------------
+
+# Comprehensive explicit map: normalized-AW-label -> canonical slug.
+# Keys are lowercased and invisible-char-stripped (as produced by
+# _normalize_fraction_label).  Slugs must match the chat's context slugs for
+# the major parties; minor/regional parties get clean derived slugs here so
+# that every one of the ~52 AW parties gets a stable, distinct slug rather than
+# falling through to "unbekannt".
+_MANIFESTO_PARTY_SLUG_MAP: dict[str, str] = {
```
romman8, line 96: this seems to be defined in multiple places. Could you move this to a single central place so that we only have to maintain it once?

ai-backend/src/ingestion/connectors/manifestos/mappers/corpus.py
```diff
@@ -0,0 +1,540 @@
...
+    extracted = trafilatura.extract(
+        html,
+        include_comments=False,
+        include_tables=True,
+        favor_recall=True,
+    )
+    return (extracted or "").strip()
+
+
+def chunk_pages(
```
romman8: same as above: please use an established chunking utility from langchain for this

ai-backend/src/ingestion/connectors/openparliament_tv/bulk.py
```diff
@@ -0,0 +1,240 @@
...
+
+        for item in session.get("data") or []:
+            if not isinstance(item, dict):
+                continue
+            if limit is not None and speeches_processed >= limit:
+                break
+
+            records = build_chunk_records(item, mdb_lookup)
+            if not records:
+                # Unaligned / ASR-only / empty-text item — D-03 skip (not counted).
```
romman8: nit: let's print/log some warning here

ai-backend/src/ingestion/connectors/openparliament_tv/client.py
```diff
@@ -0,0 +1,210 @@
...
+
+class OpTvChallengeError(RuntimeError):
+    """Raised when a pinned-host session file cannot be fetched (HTTP + curl both failed)."""
+
+
+class OpTvClient:
+    """Paced, keyless HTTP client for the openparliament.tv GitHub bulk repository.
+
+    Enforces:
+      - Sequential `sleep`-second delay between requests (keyless GitHub courtesy).
```
romman8, line 59: wdym with "keyless GitHub courtesy"?

ai-backend/src/ingestion/connectors/openparliament_tv/supersede.py
```diff
@@ -0,0 +1,232 @@
...
+    """Join (chunk_index, text) pairs in chunk_index order (C3).
+
+    A 2+-chunk speech scrolled out of order would otherwise concatenate as
+    "B+A" and SequenceMatcher("A+B", "B+A") ≈ 0.5 < 0.85 — a silent supersede
+    miss followed by a resurrection-guard re-insert.
+    """
+    return " ".join(t for _idx, t in sorted(texts, key=lambda pair: pair[0]))
+
+
+def supersede_dip_duplicates(
```
romman8, line 54: What happens if one then reruns dip ingestion after running open parliament ingestion would the dip "duplicates" reappear?

ai-backend/src/ingestion/connectors/openparliament_tv/supersede.py
```diff
@@ -0,0 +1,232 @@
...
+    # Group the just-upserted op chunks into distinct op speeches (by source_item_id),
+    # each with its speech_key and full concatenated text.
+    # ChunkRecord.source_item_id is a uuid.UUID — key by str(sid) consistently so the
+    # downstream MatchValue graft filter (which accepts only bool/int/str) never
+    # receives a raw UUID (that ValidationError silently disabled the merge forever:
+    # the per-item except fired AFTER the upsert, so the next run present-skipped).
+    op_speeches: dict[str, dict] = {}
+    for chunk in op_chunks:
+        key = getattr(chunk, "speech_key", None)
+        sid = getattr(chunk, "source_item_id", None)
```
romman8: Why do we need `getattr` to access the fields of `ChunkRecord`s?

ai-backend/src/ingestion/retrieve.py
```diff
@@ -0,0 +1,904 @@
...
+        )
+
+    if wahlperiode is not None:
+        must.append(
+            FieldCondition(key="wahlperiode", match=MatchValue(value=wahlperiode))
+        )
+
+    if legislature_period_id is not None:
+        # AW parliament_period ID — narrows to a specific legislature.
+        # NOT selective alone (same constraint as wahlperiode).
```
romman8: why do we need/want both fitering by "wahlperiode" and by "legislature_period_id"?

ai-backend/src/ingestion/retrieve.py
```diff
@@ -0,0 +1,904 @@
...
+        description=(
+            "Election region path list (e.g. ['EU', 'DE']). Chunks whose scalar "
+            "region is a member of this list are returned (VEC-03 MatchAny)."
+        ),
+    )
+    authority_tier: Optional[AuthorityTierLiteral] = Field(
+        None,
+        description=(
+            "Trustworthiness tier filter. One of: authoritative, factual_record, "
+            "self_reported, promotional. Omit to search all tiers."
```
romman8: What would be a use case where the LLM would want to set this in its tool call?

ai-backend/src/ingestion/retrieve.py
```diff
@@ -0,0 +1,904 @@
...
+    )
+    legislature_period_id: Optional[int] = Field(
+        None,
+        description=(
+            "AW parliament_period ID (e.g. 161 for 21st Bundestag, 149 for Bayern 2023-2028). "
+            "Restricts results to chunks from that specific legislature. "
+            "NOT selective alone — must be combined with source_type, party_id, or "
+            "party_ids_contains."
+        ),
+    )
```
romman8, line 860: How would the LLM doing the tool call know this id? Do we need this to be part of the tool call schema?

ai-backend/src/routes/chat.py
```diff
@@ -0,0 +1,75 @@
...
+    Returns text/event-stream with x-vercel-ai-ui-message-stream: v1 header.
+    All V1 named chat events are preserved inside `data-chat_event` parts.
+    Request body validated via Pydantic (FastAPI returns 422 on invalid input).
+    The FastAPI Request is passed to generate_chat_stream for disconnect detection.
+
+    Auth: verification is OPTIONAL (anonymous users are served normally), but
+    the body's user_is_logged_in flag (premium LLM selection) is honored only
+    when the request carries a valid `Authorization: Bearer <Firebase ID token>`.
+    """
+    body.user_is_logged_in = resolve_user_is_logged_in(
```
romman8: why is this part of the body?

ai-backend/src/routes/misc.py
```diff
@@ -0,0 +1,196 @@
...
+
+    NOTE: In the SSE model the chat history is stateless (per-request). The frontend
+    must supply the text to synthesize in the request body. V1 looked up the message by ID
+    from the socket session; that session no longer exists. The frontend holds the message
+    content and can pass it directly.
+
+    COST EXPOSURE: this endpoint synthesizes arbitrary client-supplied text and
+    is intentionally reachable without authentication (parity with anonymous
+    chat). Every request incurs a paid TTS call — if abuse appears, add rate
+    limiting and/or require the optional Bearer token (src.auth) here first.
```
romman8: Let's patch this to at least require anonymous firebase auth here

ai-backend/src/chat_service.py
```diff
@@ -22,83 +52,745 @@
...
+
+# Second-pass deep-link locator (D-02b). A cited op speech chunk carries a
+# `meta.sentence_map` (verbatim sentences + per-sentence timestamps). The locator
+# matches the model's quoted text to the best sentence and rewrites the citation
+# url to `video_uri#t={ts_start}` (phrase-level deep-link). Fuzzy fallback uses a
+# difflib ratio; below this bar the locator opens the video at the speech start.
+_DEEPLINK_FUZZY_THRESHOLD = 0.6
+
+
+def locate_deeplink(quoted: str, sentence_map: list[dict], video_uri: str) -> str:
```
romman8: why is this part of the chat service? Let's move this to another file

ai-backend/src/chat_service.py
```diff
@@ -22,83 +52,745 @@
...
+    v5 framing: ``data: <json>\n\n`` where <json> is a part object with a
+    ``type`` discriminator (start, text-delta, data-<name>, finish, ...).
+    """
+    return f"data: {json.dumps(part)}\n\n"
+
+
+def _frame(event_type: str, payload: object) -> str:
+    """Map a legacy event code to AI SDK v5 UI-message-stream part(s).
+
+    The v1 hand-rolled wire codes (f/0/8/e/d) are translated to the real v5
```
romman8: Where do these "wire codes" come from?

ai-backend/src/chat_service.py
```diff
@@ -22,83 +52,745 @@
...
+def _party_chunk(session_id: str, party_id: str, chunk: str) -> str:
+    """Emit one incremental answer chunk as a `party_chunk` data annotation.
+
+    The frontend renders the visible chat from the Zustand store, not from
+    useChat's internal `messages`. The store appends a party's answer live via
+    `mergeStreamingChunkPayloadForMessage`, which keys on `party_chunk` events
+    carrying `chunk_content` (the V1 `party_response_chunk_ready` contract).
+    The v5 `text-delta` parts feed useChat's (unrendered) message state only, so
+    without this companion event the answer only appears at `party_complete` —
+    as one block. Emitting both keeps useChat's stream valid AND streams to the UI.
```
romman8: nit: let's avoid specifiyng client implementation details in the backend code

ai-backend/src/chat_service.py
romman8: How were they chosen? This implementation seems overly complex and seems like it contains artifacts from a previous implementation

.husky/pre-commit
```diff
@@ -1 +0,0 @@
-cd web && npx lint-staged
```
SimonKaran13, line 1: Why did we delete this?

infra/cloud_run_jobs.sh
```diff
@@ -0,0 +1,103 @@
...
+#
+# One job per connector, same image, CONNECTOR_ID env var selects the connector.
+# Tasks = 1 (D-04 single-task per job; task-sharding deferred).
+# Task timeout = 3600s (1 hour — far exceeds Firebase Scheduled Functions 15-min cap;
+# Firebase's 15-min limit does NOT apply to Cloud Run Jobs which support up to 24h).
+#
+# Phase 5 connector (one surviving — others removed in D-09):
+#   abgeordnetenwatch_votes → daily (parliament votes published nightly)
+
+for CONN in abgeordnetenwatch_votes; do
```
SimonKaran13: Where and when is the `abgeordnetenwatch_votes` variable defined?

Shouldn't the script check if it's defined first?

scripts/check_gdpr_wall.py
SimonKaran13, line 1: I'm unsure of how useful this script is. I thing it makes more sense to define an instruction for @copilot to check for this kind of things when doing automatic code reviews

firebase/tests/rules.test.ts
```diff
@@ -0,0 +1,260 @@
...
+  await testEnv.cleanup();
+});
+
+describe("GDPR Art. 9 wall — users/{uid}/answers consent gate (SC3/DATA-05)", () => {
+  it("rejects answers write when consent.political_data is absent (assertFails)", async () => {
+    // Seed a user doc WITHOUT consent using Admin-equivalent bypass
+    await testEnv.withSecurityRulesDisabled(async (ctx) => {
+      await ctx
+        .firestore()
+        .doc("users/alice")
```
SimonKaran13: I couldn’t find any application code that reads from or writes to `users/{uid}/...`

firebase/tests/rules.test.ts
SimonKaran13, line 1: It's a good idea to have this kind of test, be we should check that it uses the actual collections that the app uses.

firebase/README.md
```diff
@@ -206,19 +239,21 @@ Ensure all referenced assets (logos, PDFs, etc.) exist in the prod Firebase Stor
...
 ### 4. Seed production Firestore
 
-```bash
-# From repo root:
-make seed-prod
-
-# Or from firebase/:
-ENV=prod python scripts/seed_firestore.py
-```
+Production seeding is **intentionally blocked this milestone** — the seed
```
SimonKaran13: There must be a way to seed `development` and `production` firestore. Otherwise we can't merge

firebase/README.md
```diff
@@ -206,19 +239,21 @@ Ensure all referenced assets (logos, PDFs, etc.) exist in the prod Firebase Stor
...
-The vector store collections are context-specific:
-- Dev: `context_{context_id}_party_docs_dev`
-- Prod: `context_{context_id}_party_docs`
+V2 uses a **single Qdrant collection per environment** (e.g.
+`wahlchat_chunks_dev`) holding every source for every election — NOT the V1
+per-context collections. Election scoping happens at query time via payload
+filters (`party_id` tenant key, `region`/`region_path`, `source_type`,
+`authority_tier`). See CLAUDE.md section 1 ("Qdrant Collection Design") for
+the full design. No per-context collection needs to be created when moving a
+context to prod — only the corpus chunks for the relevant region must exist.
```
SimonKaran13: Qdrant documentation should not be mixed here in the firestore directory

firebase/firestore.rules
```diff
@@ -125,8 +125,28 @@ service cloud.firestore {
...
+    match /users/{user_id}/answers/{answer_id} {
+      allow read: if request.auth != null && request.auth.uid == user_id;
+      // create/update of special-category political data requires explicit consent (Art. 9)
+      allow create, update: if request.auth != null
+        && request.auth.uid == user_id
+        && get(/databases/$(database)/documents/users/$(user_id)).data.consent.political_data == true;
+      // erasure (Art. 17) must NOT require consent — a user who has withdrawn
+      // consent must still be able to delete their own political-opinion answers
+      allow delete: if request.auth != null && request.auth.uid == user_id;
+    }
```
SimonKaran13: When is a user asked about their consent in the application? Can't we simply assume that they gave their consent by using the application? This just adds unnecessary complexity IMO

firebase/firebase.json
```diff
@@ -19,5 +19,17 @@
...
+    "firestore": {
+      "port": 8081
+    },
+    "storage": {
+      "port": 9199
+    },
+    "ui": {
+      "enabled": true,
+      "port": 4000
+    }
```
SimonKaran13, line 36: What UI is this?

SimonKaran13 (changes requested):
I tested the changes locally and the app still works. Before merging, we need to decide how to firestore will be seeded in `dev` and `prod`. We need to make sure that Qdrant is also filled with the sources. 

I like the idea of having the connectors inherit from a base class. This is a good place to have a common interface. Nonetheless, it's not very clear to me how you will deploy these in `dev` or `production`. Also I feel like we should separate the ingestion (i.e. the connectors) from the `ai-backend` code. For me it would make sense to have separate directory for this in the root of the project.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/client.py
```diff
@@ -0,0 +1,339 @@
...
+                    "(total advertised=%s, fetched=%s) — stopping pagination early.",
+                    _MAX_PAGES,
+                    path,
+                    total,
+                    len(all_data),
+                )
+                break
+
+            page_params["range_start"] = start
+            page_params["range_end"] = start + _PAGE_SIZE
```
SimonKaran13: [P1] The live API treats `range_end` as the number of requested rows, not as an absolute end index. I verified that `range_start=100&range_end=200` returns 200 rows. Because this loop requests 100, 200, 300, ... rows while advancing `start` by only 100, the pages overlap: `get_all("election-program")` returns 1,340 rows for an advertised total of 635, with only 635 unique IDs. That repeats PDF downloads and downstream fetches and can exhaust the API/job budget. Please keep `range_end=_PAGE_SIZE`, advance by the delivered row count, and update the mocked pagination tests to reflect the live contract.

ai-backend/src/routes/pro_con.py
```diff
@@ -0,0 +1,112 @@
...
+    privileged body flag, but the optional Bearer token is verified for parity
+    with /chat and /voting-behavior so future premium gating inherits it.
+    """
+    # Verified claims (or None for anonymous) — no privileged flag consumes
+    # them yet; kept so the auth contract is uniform across the SSE routes.
+    _ = verify_optional_bearer_token(request)
+
+    async def stream():
+        try:
+            party = await aget_party_by_id(body.party_id)
```
SimonKaran13: [P1] This makes both new side routes fail against the freshly seeded V2 data model. `seed_firestore.py` writes parties only to `contexts/{context_id}/parties/{party_id}`, while `aget_party_by_id()` reads the deprecated, unseeded top-level `parties/{party_id}` collection. I reproduced the pro/con and voting-behavior requests after `make seed-local`; both returned the generic error frame. Please resolve through `aget_party_for_context(context_id, party_id)`, add `context_id` to the voting-behavior DTO, and add route-level coverage using the seeded collection shape instead of mocking the legacy lookup.

ai-backend/Dockerfile
```diff
@@ -2,25 +2,46 @@
...
+# Production ingestion entrypoint (D-03 — same image, CONNECTOR_ID selects).
+# To run as the ingestion image, override CMD at docker run / job spec:
+#
+#   docker run --env CONNECTOR_ID=bundestag_votes wahlchat-ingestion
+#   gcloud run jobs create ... --set-env-vars "CONNECTOR_ID=bundestag_votes"
+#
+# The runner reads CONNECTOR_ID (or --connector arg) and dispatches to the
+# matching connector factory in src.ingestion.registry (D-03).
+#
+# Ingestion image: docker build --build-arg ENTRYPOINT=ingestion -t wahlchat-ingestion .
```
SimonKaran13: [P1] This build argument is neither declared nor used, and the only active `CMD` starts `src.app`. The Cloud Run Job spec also supplies no command override, so the deployed job starts the FastAPI server and waits until timeout instead of ever executing `src.ingestion.run`. I reproduced this from the built image. Please add a real ingestion build target/entrypoint or set the job's command and arguments explicitly, then cover the image with a smoke test that proves `CONNECTOR_ID` dispatches the runner.

infra/cloud_run_jobs.sh
```diff
@@ -0,0 +1,103 @@
...
+  # Job NAME must match Cloud Run's [a-z]([-a-z0-9]*[a-z0-9])? — no underscores,
+  # so connector-id underscores become hyphens (G4). CONNECTOR_ID env var keeps
+  # the underscore form: it is the registry key the runner resolves.
+  #
+  # `deploy` (not `create`) is idempotent: it creates the job on first run and
+  # updates it on re-runs, so this script can be re-applied safely (G18).
+  gcloud run jobs deploy "wahlchat-${CONN//_/-}" \
+    --image "${IMAGE}" \
+    --region "${REGION}" \
+    --set-env-vars "CONNECTOR_ID=${CONN},ENV=prod,QDRANT_URL=${QDRANT_URL}" \
```
SimonKaran13: [P1] Even after fixing the image command, this job cannot safely target the intended corpus. The script defaults to `PROJECT=wahl-chat-prod`, but the actual production project ID is `wahl-chat`; overriding it to `wahl-chat-dev` still hard-codes `ENV=prod`. It also supplies neither `QDRANT_API_KEY` nor `AW_LEGISLATURE_ID`, so a protected store rejects the job and the connector otherwise falls back to legislature 111 (the 19th Bundestag), leaving the current Bundestag and Landtage unpopulated. Please require and validate the project/environment pair, inject the Qdrant secret, and create explicitly scoped jobs for the intended legislature IDs.

README.md
```diff
@@ -21,16 +21,22 @@ The aim of wahl.chat is to enable users to engage in a contemporary way with the
...
-- [Firebase CLI](https://firebase.google.com/docs/cli) (for `firebase/`)
+- [Node.js](https://nodejs.org/) 22+
+- [bun](https://bun.sh/) 1.3.x — install via Homebrew: `brew install bun`, or `curl -fsSL https://bun.sh/install | bash`
+- [uv](https://docs.astral.sh/uv/) 0.11.x — `curl -LsSf https://astral.sh/uv/install.sh | sh`
+- [Docker](https://www.docker.com/) 29+ with Compose v5+
+- [Firebase CLI](https://firebase.google.com/docs/cli) 15.x — `npm install -g firebase-tools`
+
+## Local development
+
+No production credentials needed. All data stores run locally.
```
SimonKaran13, line 39: [P1] The web application is not connected to the local Firebase emulators. The Make target starts only Firestore, neither browser/server Firebase initialization calls `connectFirestoreEmulator` or `connectAuthEmulator`, and no Auth emulator is configured. With dev Firebase values, a local browser therefore signs in anonymously and writes to the real dev project while FastAPI reads the emulator; with the blank example values, Firebase cannot initialize. Please add an explicit local-emulator switch for both SDKs, configure/start Auth and Firestore in `dev-local`, and only then claim that all stores run locally.

ai-backend/src/auth.py
```diff
@@ -0,0 +1,71 @@
...
+        return False
+    claims = verify_optional_bearer_token(request)
+    if claims is None:
+        logger.debug(
+            "%s: request body claims user_is_logged_in but presented no valid "
+            "Firebase ID token — ignoring the flag (treated as anonymous).",
+            route,
+        )
+        return False
+    return True
```
SimonKaran13, line 95: [P1] A valid Firebase token does not prove that this is a registered user: every ordinary visitor is automatically signed in anonymously and receives a valid token. Such a caller can change `user_is_logged_in` to `true`, present that anonymous token, and enable the premium-model branch. Please reject `claims["firebase"]["sign_in_provider"] == "anonymous"` or, preferably, require a server-controlled entitlement/custom claim, and add an anonymous-token regression test.

ai-backend/src/chat_service.py
```diff
@@ -466,376 +1518,692 @@ async def process_party(
...
+            continue
+        if not first_user_seen:
+            first_user_seen = True
+            continue
+        preceding = chat_history[idx - 1] if idx > 0 else None
+        if preceding is None or preceding.role != Role.ASSISTANT:
+            return False
+        if not preceding.quick_replies:
+            return False
+        if msg.content not in preceding.quick_replies:
```
SimonKaran13: [P1] This does not establish that a follow-up was curated because the entire `chat_history` - including assistant messages and their `quick_replies` - comes from the request body. A client can fabricate an assistant turn whose quick replies contain arbitrary political text, repeat that text as the next user turn, and make this predicate return `true`; the resulting conversation is then eligible for the shared cache despite the GDPR rationale above. Please disable multi-turn caching until quick-reply provenance can be verified server-side (for example with signed opaque IDs), and add a forged-history regression test.

ai-backend/.dockerignore
```diff
@@ -5,3 +5,12 @@
 data/wahlprogramme/
 firebase/
 .venv/
+# Secrets must NEVER enter the build context (COPY . /app would bake them).
+.env
+.env.*
```
SimonKaran13, line 10: [P1] Service-account files still enter the Docker build context. The backend README explicitly tells Docker/CI operators to place `wahl-chat-dev-firebase-adminsdk.json` or `wahl-chat-firebase-adminsdk.json` in this directory, and `COPY . /app` then bakes that private key into both image layers and the runtime image; `.gitignore` does not protect Docker contexts. Please exclude `*firebase-adminsdk*.json` (and ideally all credential JSON patterns) here and use workload identity or a runtime-mounted secret instead of copying credentials into the source tree.

Makefile
```diff
@@ -10,7 +15,34 @@ install-web:
 	cd web && bun install
 
 install-backend:
-	cd ai-backend && poetry install
+	cd ai-backend && uv sync
+
+# --- Local data stores (LOCAL-04) ---
+
+stores-up:
+	docker compose up -d qdrant
```
SimonKaran13, line 36: [P1] A fresh Qdrant volume does not contain `wahlchat_chunks_dev`, but none of the documented local steps invokes the idempotent `src.ingestion.setup_collection`. The runner calls `get_cursor()` before creating anything, so I reproduced the first connector run failing with a 404 for the missing collection. Please wait for Qdrant readiness and run/verify the schema setup here (or in an explicit bootstrap target used by `dev-local`), with an equivalent bootstrap/migration step for dev and production deployments.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,250 @@
...
+        working-directory: .
+        run: python3 scripts/check_gdpr_wall.py
+
+      - name: Lint — ruff
+        run: uv run --no-sync ruff check src/
+
+      - name: Type-check — mypy
+        run: uv run --no-sync mypy src/
+
+      - name: Unit / integration tests (excluding smoke test)
```
SimonKaran13, line 127: [P2] The Qdrant schema module skips wholesale when localhost Qdrant is unavailable, and this job never starts Qdrant, so those integration tests are absent from every green CI run. With Qdrant running locally, the suite currently fails because `test_new_phase5_indexes_present` still requires the removed `status`, `as_of_date`, and `claim_id` indexes. Please provision a Qdrant service for this job, make unavailability fail rather than skip, and derive the assertion from the canonical schema/index specification.

ai-backend/src/routes/voting_behavior.py
```diff
@@ -0,0 +1,292 @@
...
+                party, body.last_user_message, body.last_assistant_message
+            )
+
+            # Retrieve vote_record chunks from the single wahlchat_chunks store.
+            # source_type='vote_record' is the selective filter that keeps this cheap
+            # on the m=0 tenant-only HNSW (see module docstring re: full_scan_threshold).
+            vote_payloads: list[dict] = await asyncio.to_thread(
+                retrieve,
+                improved_rag_query,
+                source_type="vote_record",
```
SimonKaran13, line 241: [P1] This selects the global top 10 vote records before checking whether the requested party participated. For a smaller party, all ten candidates can be discarded even though relevant party votes exist immediately below them; the missing context filters can also mix Bundestag and Landtag records. Please apply `party_ids_contains=party.party_id` inside `retrieve()` before top-k selection and thread the context's region/legislature/election level into the query.

ai-backend/src/routes/voting_behavior.py
```diff
@@ -0,0 +1,292 @@
...
+            url=citation_url,
+            date=date_str,
+            title=title,
+            subtitle=outcome,
+            detail_text=payload.get("text"),
+            links=links,
+            voting_results=VotingResults(overall=overall, by_party=by_party),
+            short_description=outcome,
+            vote_category=None,
+            submitting_parties=list(payload.get("party_ids") or []),
```
SimonKaran13: [P1] `party_ids` is defined by the ingestion mapper as every party participating in the roll call, but this field is rendered by the UI as "Einreichende Partei(en)." This will falsely identify essentially every voting faction as a sponsor of the motion. Please leave this unset/hide the field until actual submitter metadata is ingested, or introduce a distinct `submitting_party_ids` field sourced from the motion.

ai-backend/src/routes/chat.py
```diff
@@ -0,0 +1,75 @@
...
+}
+
+
+class ChatRequestDto(BaseModel):
+    session_id: str
+    context_id: str
+    user_message: str
+    party_ids: list[str]
+    user_is_logged_in: bool = False
+    chat_history: list = []
```
SimonKaran13: [P1] The SSE migration drops the server-side limits enforced by `ChatUserMessageDto`, including the 500-character user-message bound, and accepts an unbounded, untyped client-supplied history and party list. Because this endpoint is public and these values feed directly into LLM prompts, callers can bypass the UI and submit arbitrarily large histories, causing avoidable model cost, context overflow, and memory pressure. Please use constrained, typed Pydantic fields (including a bounded `list[Message]`, bounded party count, and `default_factory=list`) and add regression tests asserting that invalid payloads receive HTTP 422 responses.

SimonKaran13 (changes requested):
I reviewed the changed paths end to end and focused on issues that were not already covered by the existing review threads.

I am requesting changes because the current branch has several functional and security blockers: the scheduled container never invokes ingestion, fresh Qdrant cannot be used by the documented local flow, local browser traffic is not connected to the Firebase emulators, anonymous callers can opt into premium models, and the GDPR cache gate trusts client-forgeable history. I also found data-correctness problems in live Abgeordnetenwatch pagination and the voting-behavior path.

Validation performed: frontend lint, type-check, and production build passed; Firebase rules tests passed (12/12); Ruff, mypy, the SSE smoke tests, and local-mode guard passed; 564 backend tests passed when the live Qdrant schema module was excluded. When Qdrant is available, the suite exposes a stale failing schema test that CI currently skips. I also reproduced the Cloud Run image command and the Abgeordnetenwatch pagination behavior directly.

Detailed, actionable findings are inline.

.github/workflows/ci.yml
annikasiefke: split it up!

.github/workflows/ci.yml
annikasiefke: At work we run on every push, so you don't need to open PRs only to run the pipeline. But I changed it here to run only on PRs and pushes to develop and main.

.github/workflows/ci.yml
annikasiefke: Fixed the test and updated CI as suggested (Qdrant service, failure on unavailability and derived assertions)

.husky/pre-commit
annikasiefke: Added it back, but for me the combination of .husky pre-commit and pre-commit-config didn't work locally, only the husky pre-commit ever ran. Not sure if I am missing some local set up for that to work, but running pre-commit from inside here fixed it for me, so both types of hooks can run.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/stance.py
annikasiefke: Party slugs are all political-domain terms and everything is German, so I'd maybe keep it in German. But let me know, can change it as well!

ai-backend/src/ingestion/connectors/bundestag_speeches/client.py
annikasiefke: Done as suggested

ai-backend/src/ingestion/connectors/bundestag_speeches/client.py
annikasiefke: DIP documents a concurrency cap (25 concurrent), so the 0.2 s inter-request sleep is deliberate pacing to stay a well-behaved client on a keyed public API, not a rate-limit workaround. So would keep it, what do you think?

ai-backend/src/ingestion/connectors/bundestag_speeches/mdb.py
annikasiefke: The speeches reference politicians and generally contain a <fraktion> tag, but this is not always there, so we need to use this to map politicians to their party at the point of the speech. Removed the comment about "Absorbed..."

ai-backend/src/ingestion/connectors/bundestag_speeches/parser.py
annikasiefke: I kept this for now, because rede_id is the exact xml key used. Would you still rename it?

ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
annikasiefke: A full Bundestag term has a few hundred polls (each poll = one recorded vote); the Landtage have fewer.

ai-backend/src/ingestion/connectors/bundestag_speeches/constants.py
annikasiefke: Added a `make fetch-mdb-stammdaten` target to download it in the correct dir

ai-backend/src/ingestion/connectors/manifestos/mappers/corpus.py
annikasiefke: One standard one in `src/ingestion/party_slugs.py` (major-party core + state/regional layer + shared normalizer) -> each connector composes its map from it.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/stance.py
romman8: fair, fine to keep as is

ai-backend/src/routes/chat.py
```diff
@@ -0,0 +1,88 @@
...
+_MAX_HISTORY_MESSAGES = 100
+
+
+class ChatRequestDto(BaseModel):
+    session_id: str = Field(max_length=200)
+    context_id: str = Field(max_length=200)
+    user_message: str = Field(max_length=_MAX_USER_MESSAGE_CHARS)
+    party_ids: list[str] = Field(default_factory=list, max_length=_MAX_PARTY_IDS)
+    user_is_logged_in: bool = False
+    chat_history: list[Message] = Field(
```
Antonwy: Why is chat_history considered bounded when only the list length is capped? Message.content and the nested output fields remain unbounded, so one unauthenticated request can still allocate a huge body and send a huge prompt to the LLM. Please use a route-specific role/content input type, bound each turn plus total history size, and test one oversized nested turn.

ai-backend/src/routes/misc.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+    if verify_optional_bearer_token(request) is None:
+        raise HTTPException(status_code=401, detail="Firebase authentication required")
+    try:
+        # Frontend sends the text to synthesize in the voice field (repurposed as text)
+        # The TTS endpoint requires the text content to be provided
+        # via a separate TtsRequestDto. We accept TextToSpeechRequestDto as-is since
+        # the text is embedded in voice field; the audio service synthesizes any text.
+        # This interface can be cleaned up later.
+        text_for_speech = sanitize_text_for_speech(body.voice)
```
Antonwy: Why does this endpoint now synthesize arbitrary caller-owned text? The old socket path looked up a server-owned message; this route accepts the voice field as text and gates it only on any valid Firebase token, including the anonymous tokens issued to visitors. With no ownership check, input limit, rate limit, or quota, this exposes a paid TTS proxy. I would remove the dormant endpoint or restore server-owned lookup with authorization and limits.

ai-backend/src/ingestion/run.py
```diff
@@ -0,0 +1,678 @@
...
+                # surfaced by an AW_REFRESH reconcile run) is re-written rather than
+                # silently skipped. Connectors that don't stamp content_hash keep plain
+                # point-id-existence idempotency (content_changed stays False for them).
+                point_ids = [
+                    str(compute_chunk_id(c.source_item_id, c.chunk_index))
+                    for c in chunks
+                ]
+                new_point_ids = set(point_ids)
+                # source_item_id is a UUID; stringify for the keyword-index MatchAny.
+                siids = sorted({str(c.source_item_id) for c in chunks})
```
Antonwy, line 453: How does this remove a child that disappears completely from a multi-speech parent? The footprint is built only from source_item_ids still emitted by the new normalize call. If a protocol previously produced A+B and now produces only A, B is never scanned and remains retrievable forever. Please store a stable connector-parent identity, reconcile the full old parent footprint, and add an A+B to A regression test.

Antonwy:
I left three inline blockers:

- nested chat history remains unbounded despite the list-count cap;
- the new TTS route accepts arbitrary paid input from anonymous users;
- ingestion cannot reconcile a whole child speech that disappears from a multi-speech parent.

After those, I would also address:

- include the connector source in parliamentary source_item_id generation so DIP and op identities cannot collide;
- extract the duplicated single-party/comparison retrieval orchestration from chat_service.py;
- add an explicit rollout gate or deployment dependency before switching the runtime to a corpus whose scheduled ingestion and bootstrap are still deferred;
- replace inspect.getsource and obsolete capability-guard tests with behavioral coverage;
- remove the remaining FIX, Rule, Phase, CR, and PR review references from comments.

The underlying direction is good: the frozen ingestion envelope, content hashes, embed-before-delete ordering, dimension guard, real-Qdrant CI, auth hardening, emulator wiring, Docker secret exclusions, and PDF proxy validation are all solid. The current head is 30ecf31c and all checks are green; backend CI reports 564 passed and 2 skipped.

Antonwy:
One way I would split this into a dependency-ordered stack:

1. **Tooling baseline** — Poetry to uv, lockfiles, pre-commit, Docker, and CI. No runtime behavior change.
2. **Transport migration** — aiohttp and Socket.IO to FastAPI and SSE, plus the frontend adapter, while preserving the existing retrieval behavior. This isolates protocol compatibility.
3. **Corpus foundation** — ChunkRecord, deterministic IDs, Qdrant collection and indexes, generic runner, registry, and real-Qdrant tests. No connectors or chat cutover yet.
4. **Votes** — Abgeordnetenwatch connector, vote mapping, scoped retrieval, and voting-behavior integration.
5. **Manifestos** — connector, parsing and page metadata, retrieval, and citation contract.
6. **DIP speeches** — connector, speaker and party resolution, retrieval, and PDF citations.
7. **openparliament enrichment** — bulk connector, cross-source dedup and supersede behavior, sentence timing, and video deep links.
8. **Citation UI** — sources dialog, PDF proxy, and media viewer against a fixture-backed API contract.
9. **Rollout and cutover** — Terraform jobs and scheduler, secrets and Workload Identity, collection bootstrap and snapshot, Firestore seed, then the feature-flagged switch to the new corpus.

I would use gh stack because each later PR depends on contracts introduced earlier. Keep the transport behind a compatibility adapter and make the final rollout PR the only one that changes the production data source. That gives every step a focused review, an independently testable contract, and a clear rollback.

ai-backend/src/ingestion/connectors/openparliament_tv/supersede.py
annikasiefke: DIP's `normalize()` runs a cursor-independent resurrection guard (`is_op_superseded()`): before building each chunk it scrolls Qdrant for the `op` twin by `speech_key`, text-matches (ratio ≥ 0.85) and **skips the DIP speech entirely**. The only constraint is that op and dip ingestion should not run in parallel.

infra/cloud_run_jobs.sh
annikasiefke: I removed this script for now since we plan to do a proper infra set up with terraform for the deploy -> can define proper cloud run jobs and env set up for connectors there. Would postpone that from this MR since it is already quite large and we can for now just migrate a snapshot of local Qdrant to our deployed qdrant or run ingestion with the make targets pointed at the remote sources instead of local ones.

firebase/README.md
annikasiefke: Added a guarded seed path: `seed_firestore.py` gains a `SEED_TARGET=real` mode (refuses without an explicit project, refuses prod without a confirmation) that seeds only the config collections (contexts / parties / proposed_questions). Would propose to have a GitHub action to seed this when we do the terraform set up, so firestore is auto-seeded whenever new election contexts are added.

scripts/check_gdpr_wall.py
annikasiefke: I thought having it is nice, since GDPR is legally binding and like this we have a fail save check. But we can/ should still add instructions for copilot to check for these kind of things during automatic code reviews. Is this something I need to do in the code or do we define it somewhere here on github?

ai-backend/src/ingestion/connectors/openparliament_tv/client.py
annikasiefke: We here fetch openparliament.tv's data from their public GitHub repo without an API token (keyless). Unauthenticated GitHub is rate-limited, so we add a 0.2 s inter-request sleep to stay polite and avoid throttling.

ai-backend/src/ingestion/retrieve.py
annikasiefke: dropped the redundant wahlperiodee filter

firebase/firestore.rules
annikasiefke: As far as I understand we can't. Political opinions are special category data according to GDPR Article 9 and cannot be processed without explicit consent from the user (https://gdpr-info.eu/art-9-gdpr/). To me simply using the application is not explicit consent and this sounds like it also requires a more clear form of giving consent: https://gdpr-info.eu/recitals/no-32/. I believe storing the chat messages already qualifies as processing the data. At the very latest this will be needed once we use the data for user alignment scoring, but how I understand this, it is already required simply by storing it. 
Removed this part for now, since we're not asking for consent in anywhere, but I think we should be asking for consent.

firebase/firebase.json
annikasiefke: It's the Firebase Emulator Suite UI, so the local dashboard the Firebase CLI serves at http://localhost:4000 when the emulators are running (`make stores-up`). It's a browser console for inspecting/editing the emulated Auth (:9099), Firestore (:8081), and Storage (:9199) data during local dev.

.gitignore
annikasiefke: Not just the ai-backend uses python, other modules have python script as well, added a comment to `.gitignore`.

ai-backend/src/ingestion/connectors/bundestag_speeches/mappers/corpus.py
annikasiefke: Our original idea was to have one chunk per data point hence the 6000 tokens. Moved everything to one shared boundary-aware chunker (`src/ingestion/chunking.py`) wrapping langchain's `RecursiveCharacterTextSplitter`, with per-connector overrides, copying how you previously chucked the pdf Wahlprogramme.

infra/cloud_run_jobs.sh
annikasiefke: see my comment above about removing this one for now.

README.md
annikasiefke: Added local emulators for both SDKs.

ai-backend/src/chat_service.py
annikasiefke: Restored develop's model. The session the web socket used was dropped, but now we at least have a small server-side session store (in-memory, keyed by `session_id`) that holds the quick replies + cacheability and the gate checks that, never client history. Also "missing state → not cacheable", so a cold start costs only a cache hit, never a leak.

ai-backend/src/routes/chat.py
annikasiefke: I fixed this, but also noticed that develop's `ProConPerspectiveRequestDto`, `VotingBehaviorRequestDto` and `ParliamentaryQuestionRequestDto` (ai-backend/src/models/dtos.py on develop) all declare these as bare `str = Field(...)` with no `max_length`. Should these also get limits?

Makefile
annikasiefke: **Local**: Added new idempotent `make bootstrap-collection` target (`python -m src.ingestion.setup_collection`) is a prerequisite of `dev-local` and of every ingestion `run-*` target and it runs only after `stores-up` has polled Qdrant's /readyz. So the first `get_cursor()` no longer 404s on a fresh volume.

**Dev/prod**: The deployable image's entrypoint runs the same idempotent `setup_collection` before the connector whenever `CONNECTOR_ID` is set (`ai-backend/docker-entrypoint.sh`)

ai-backend/src/routes/chat.py
annikasiefke: Fixed as suggested

ai-backend/src/routes/misc.py
annikasiefke: removed the dormant endpoint

ai-backend/src/ingestion/run.py
annikasiefke: Thanks for pointing this out, fixed as suggested

PR #113, opened 2026-07-25
V2 stack 0/9: Clear V1 application code (stack base)

annikasiefke:
Base of the V2 stack — this splits #111 into dependency-ordered, individually reviewable PRs. Each PR chains off the previous one; merging the whole stack produces a tree byte-identical to `v2-migration`.

This PR is **pure deletion**: it clears the V1 application code from this branch line so V2 can be reviewed as what it is — a rebuilt system — in nine additive, individually green PRs, instead of a 55k-line mixed diff. V2 forked V1's frontend but rebuilt the data layer and backend around a new corpus, so a file-by-file evolution diff mostly obscures what changed; #111 remains available for anyone who wants that view.

Nothing about any deployed environment changes until the stack merges (same end state as merging #111). V1 stays deployed from its current branches; rollback is a revert.


annikasiefke:
Superseded: the stack was rebuilt as a progressive diff off develop — no V1-removal base needed. The V1→V2 deletions now land in the layers that replace them (transport PR removes the aiohttp/Socket.IO app, frontend PR removes the socket client, rollout PR removes the München seed data).

PR #114, opened 2026-07-25
V2 stack 1/9: Tooling baseline

annikasiefke:
First PR of the V2 stack (1/9) — this splits #111 into dependency-ordered, individually reviewable PRs. Each PR chains off the previous one and shows the progressive V1→V2 diff for its subsystem; merging the whole stack produces a tree byte-identical to `v2-migration`.

This one is tooling only, no runtime behavior change:

- **Poetry → uv**: `poetry.lock` removed; `uv.lock` + the uv-based flow arrive with the backend packaging in the next PR.
- **Makefile** — the entry point for all workflows (install, local stores, dev servers, tests, ingestion runs).
- **docker-compose** for local Qdrant; updated pre-commit + husky hooks.
- **`scripts/check_gdpr_wall.py`** — the CI guard that fails the build if ingestion code ever references Firestore `users/` (GDPR Art. 9 wall).
- **AGENTS.md** as the single source-of-truth repo guide (provider files like CLAUDE.md become symlinks via `scripts/setup-agent-docs.sh`), plus licensing housekeeping and `.cursorignore`/`.DS_Store` cleanup.

One deliberate omission: the updated CI workflow lands in the last PR of the stack — its jobs run against the V2 backend/frontend/firebase trees, which take shape over the following PRs.


.pre-commit-config.yaml
```diff
@@ -1,32 +1,39 @@
...
     hooks:
-      - id: ruff
-        name: Ruff (linting)
-        args: ["--fix"]
+      - id: ruff-check
+        name: Ruff (lint)
+        entry: uv run --project ai-backend ruff check --fix
+        language: system
         files: ^ai-backend/.*\.py$
+        require_serial: true
```
romman8, line 18: What is the reason for setting `require_serial` here?

AGENTS.md
```diff
@@ -0,0 +1,344 @@
...
+- **backend** — GDPR wall guard → `uv sync --locked` → ruff → mypy → pytest
+  (excluding the smoke and local-mode tests).
+- **frontend** — `bun install --frozen-lockfile` → `bun audit --audit-level=high`
+  (supply-chain gate) → lint → tsc → build.
+- **firebase-rules** — Firestore security-rules tests against the emulator
+  (proves the chat-message privacy rules).
+- **smoke-test** — the E2E SSE smoke test, in-process via `httpx.ASGITransport`
+  (needs backend + frontend green first).
+
+Before opening a PR, run the lint / type-check / test commands for whichever
```
romman8: Imo running linting should not be an explicit instruction for the agent but rather covered by the tooling (e.g. pre-commit) so that the agent does not bother with things that will be run anyways when it commits. The agentic loop should rather be focused on verifying its own changes with existing regression tests and new integration and/or unit tests

AGENTS.md
```diff
@@ -0,0 +1,344 @@
...
+    fresh.
+- **No internal references, anywhere.** Internal planning/decision/tracking
+  identifiers must never appear in the codebase — not in comments, and not in
+  runtime strings (error messages, logs, `argparse` help, `Field(description=)`,
+  printed output). This covers decision/plan/research IDs (`D-…`, `T-…`, `C…`,
+  `G…`, `SC…`, `VEC-…`, `FOUND-…`), phase numbers (`Phase 9`), plan/task refs
+  (`plan 01-08`, `Task 2`), doc filenames (`RESEARCH.md`, `PLAN.md`), and
+  code-review finding IDs (`HIGH-1`, `MED-02`). Write the reasoning or the
+  user-facing message, never the paper trail.
+- **Package managers.** uv for `ai-backend`, bun for `web`. Add Python deps with
```
romman8: Can drop the last 3 points describing tooling as they are already obvious from the setup and rest of the doc

AGENTS.md
```diff
@@ -0,0 +1,344 @@
...
+Provider-specific files (e.g. `CLAUDE.md`) are symlinks to this file, created by
+`./scripts/setup-agent-docs.sh`. Edit `AGENTS.md` — never a symlink.
+
+## Overview
+
+wahl.chat is a grounded political information chatbot for German elections
+(federal, state, and local). It answers questions about parties' positions and
+substantiates its answers with citations to primary sources.
+
+wahl.chat V2 forks the V1 application largely as-is and concentrates new work on
```
romman8: This will age pretty poorly --> would drop the V2 vs V1 framing and would rather describe the status-quo of v2

scripts/setup-agent-docs.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+# SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
+#
+# Link provider-specific agent-instruction files to the committed source of truth.
+#
+# AGENTS.md is the single, committed agent/onboarding document. Some assistants
+# (e.g. Claude Code) look for a provider-specific CLAUDE.md instead, so we point
+# it at AGENTS.md via a symlink. The symlink is git-ignored — only AGENTS.md is
+# committed — so this script is how each clone materialises CLAUDE.md locally.
```
SimonKaran13, line 9: Isn't it easier to just have the symlink commited? Then there is no need to have this script and Claude would see the CLAUDE.md out of the box.

Makefile
```diff
@@ -10,7 +15,80 @@ install-web:
 	cd web && bun install
 
 install-backend:
-	cd ai-backend && poetry install
+	cd ai-backend && uv sync
```
SimonKaran13, line 18: [P1] `pyproject.toml` still only defines a Poetry project and this PR adds no `uv.lock`. On this exact head, this command exits 0 but installs zero packages; `uv tree` then reports “No project table found”, and the new test/pre-commit targets cannot spawn pytest or pre-commit. The same Makefile targets `src.app`, SSE tests, and ingestion modules that are not in this PR. Either bring the uv/runtime conversion into this stack item or keep the current Poetry/V1 commands until their dependencies land.

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      - "6333:6333"
+      - "6334:6334"
+    volumes:
+      - qdrant_storage:/qdrant/storage
+    environment:
+      QDRANT__SERVICE__GRPC_PORT: 6334
+
+  # Ingestion runs the SAME image that deploys as a Cloud Run Job — the entrypoint
+  # dispatches on CONNECTOR_ID (see ai-backend/docker-entrypoint.sh), so running it
+  # here exercises the real artifact (locked deps, collection-bootstrap guard)
```
SimonKaran13, line 14: [P1] This service cannot build on this stack item. The unchanged Dockerfile still runs `COPY pyproject.toml poetry.lock /app/`, but this PR deletes `poetry.lock`; it also references `docker-entrypoint.sh`, which is not present. I reproduced `docker build ai-backend` failing at the missing lockfile. Include the Docker packaging/entrypoint changes here or move this service to the PR that introduces them.

README.md
```diff
@@ -21,40 +21,95 @@ The aim of wahl.chat is to enable users to engage in a contemporary way with the
...
 ### 2. Configure environment variables
 
 ```bash
-cp web/.env.example web/.env.local
 cp ai-backend/.env.example ai-backend/.env
+cp web/.env.example web/.env.local
 ```
 
-See sub-project READMEs for details on required variables.
+The `.env.example` files are pre-configured for local mode (`QDRANT_URL=http://localhost:6333`, `FIRESTORE_EMULATOR_HOST=localhost:8081`). You only need to fill in LLM API keys (`OPENAI_API_KEY`, `GOOGLE_API_KEY`) for the chat to work.
```
SimonKaran13, line 58: [P1] Neither example file contains the local configuration described here on this head: `ai-backend/.env.example` still has `QDRANT_URL=your_qdrant_url`, `web/.env.example` has no emulator switch/hosts, and `firebase/firebase.json` has no Firestore port 8081. Following the new README therefore does not produce an isolated local app. Please land the documentation and Make targets together with the supporting Firebase/env changes, or keep the current instructions in this PR.

Makefile
```diff
@@ -10,7 +15,80 @@ install-web:
...
+# Override these to target a deployed store (there is no scheduled orchestration
+# yet), e.g.:
+#   make run-speeches QDRANT_URL=https://<host>:6333 QDRANT_API_KEY=<key> ENV=prod
+# ENV scopes the collection name (wahlchat_chunks_{ENV}); dev and prod are isolated.
+QDRANT_URL ?= http://localhost:6333
+ENV ?= dev
+# QDRANT_API_KEY is unset by default (local Qdrant is unauthenticated) and is only
+# forwarded when non-empty, so an empty string never reaches the client as a bogus
+# credential. QDRANT_ENV is the env prefix every ingestion recipe runs under.
+QDRANT_ENV := QDRANT_URL=$(QDRANT_URL) ENV=$(ENV) $(if $(QDRANT_API_KEY),QDRANT_API_KEY=$(QDRANT_API_KEY))
```
SimonKaran13: [P1] Make prints the expanded recipe before running it, so the documented `make … QDRANT_API_KEY=<key>` invocation writes the production key into terminal or CI logs; `make -n` shows the full value. Please require the key through an already-exported environment/secret file and keep it out of rendered command lines.

.husky/pre-commit
```diff
@@ -1 +1,10 @@
-cd web && npx lint-staged
+set -e
+
+# Frontend (web/): ESLint + Biome on staged JS/TS.
+(cd web && npx lint-staged)
```
SimonKaran13: [P2] `AGENTS.md` now makes bun the frontend package manager, but this hook still uses `npx`. That keeps a second package-manager path and can let npx fetch a different `lint-staged` version when the local binary is absent. Please run the locked dependency through bun as well.

SimonKaran13 (changes requested):
I am requesting changes because this stack item is not a working tooling baseline on its own. The dependency order is inverted: the new commands, documentation, and Compose service point at packaging, runtime modules, tests, ingestion code, and Firebase configuration that only arrive in later PRs.

On the exact PR head I reproduced: `make install-backend` exits successfully but installs no backend dependencies; `uv tree` reports that there is no project table; `make test-backend` cannot spawn pytest; `make dev-backend` cannot find `src.app`; the GDPR guard fails because the ingestion tree does not exist; and `docker build ai-backend` fails because the Dockerfile still copies the deleted `poetry.lock`. The green checks are Vercel previews only, so none of these backend paths are covered.

The new dependency pins are also already behind the current releases: Qdrant is 1.18.3 and pre-commit-hooks is 6.0.0.

Either keep this PR compatible with the current Poetry/aiohttp tree, or move the uv packaging, lockfile, Dockerfile, Firebase-local configuration, and relevant runtime targets into this first stack item so it is independently usable. Detailed issues are inline.

.pre-commit-config.yaml
annikasiefke: Without `require_serial`, pre-commit chunks the staged file list and runs the entry multiple times in parallel, which means concurrent `uv run` bootstraps racing on the same environment and `ruff --fix` writing files from several processes at once. Serial keeps one uv invocation per hook.

scripts/setup-agent-docs.sh
annikasiefke: I think committed symlinks break on Windows checkouts unless `core.symlinks`/developer mode is enabled, that's why I added the script. If everyone is using macOS/Linux we can also commit it.

Makefile
annikasiefke: Accepted as is for now.

README.md
annikasiefke: Accepted as is for now.

docker-compose.yml
annikasiefke: Accepted as is for now.

SimonKaran13 (approved):

PR #115, opened 2026-07-25
V2 stack 2/9: Corpus foundation — data contract, runner, retrieval, embeddings factory

annikasiefke:
V2 stack 2/9 — the data layer's common ground: everything a connector needs to exist, before any connector does.

- **`ChunkRecord` data contract** (`src/ingestion/schemas.py`, frozen Pydantic, `extra="forbid"`) with the `AuthorityTier`/`SourceType` enums and typed `VoteMeta`/`SpeechMeta` builders.
- **One Qdrant collection per environment** (`wahlchat_chunks_<env>`): `party_id` as tenant key (no global HNSW graph), scalar `region` + payload indexes, 3072-dim COSINE locked in `setup_collection.py`.
- **Generic runner** (`run.py`): cursor derived from Qdrant itself (no watermark store), deterministic point IDs + `content_hash` for idempotent re-runs, parent-footprint reconciliation for shrunken documents, batch/time budgets for the 15-minute scheduled-job cap.
- **`BaseConnector`** — the pure `discover → fetch → normalize` interface every connector implements.
- **Filtered retrieval** (`retrieve.py`): party/region/source-type/date scoping, two-pass current+historic support.
- **Embeddings provider factory** (`src/embeddings.py`): OpenAI `text-embedding-3-large` default, Gemini selectable via env (`EMBEDDING_PROVIDER`/`EMBEDDING_MODEL`/`EMBEDDING_DIM`), ingest/query `task_type` wiring. It lives here rather than in a later PR because the runner and retrieval import it at module level.

The connector registry ships here with deferred (function-body) imports — the concrete connectors arrive in the next PRs.

Verified: ruff, mypy, 88 tests, GDPR wall guard.


ai-backend/src/ingestion/chunking.py
```diff
@@ -0,0 +1,78 @@
...
+        return []
+    return make_splitter(chunk_size, chunk_overlap).split_text(text)
+
+
+def chunk_pages(
+    pages: list[tuple[int, str]],
+    chunk_size: int = DEFAULT_CHUNK_SIZE,
+    chunk_overlap: int = DEFAULT_CHUNK_OVERLAP,
+) -> list[tuple[str, int, int]]:
+    """Split page-annotated text, keeping each chunk within a single page.
```
romman8, line 60: What is your rationale for this?
For PDFs like manifestos we often want a chunk to span across a page boundary as we would otherwise split text that clearly belongs together and simply breaks across the page boundary into separate chunks.

See e.g. page 18-19 in https://www.spd.de/fileadmin/Dokumente/Beschluesse/Programm/SPD_Programm_bf.pdf:

Image

romman8: I see that our original usage of `text_splitter.split_documents(pages)` had this issue aswell --> we can keep as is for now. Would be great to improve this in a follow-up though

ai-backend/src/embeddings.py
```diff
@@ -0,0 +1,115 @@
...
+    if resolved_provider == "gemini":
+        # AI Studio access: GOOGLE_API_KEY is the primary name (matches llms.py);
+        # GEMINI_API_KEY is accepted as an alias. output_dimensionality pins the
+        # vector width to the collection so run.py's dimension guard passes.
+        from langchain_google_genai import (  # noqa: PLC0415
+            GoogleGenerativeAIEmbeddings,
+        )
+
+        api_key = os.getenv("GOOGLE_API_KEY") or os.getenv("GEMINI_API_KEY")
+        return GoogleGenerativeAIEmbeddings(
```
romman8, line 105: https://ai.google.dev/gemini-api/docs/embeddings#task-types-embeddings-2

Does this take care of the new task type specific input formatting internally? If no, we should make sure we use the query/document structure suggested by Google when we use `gemini-embedding-002`

ai-backend/src/ingestion/__init__.py
```diff
@@ -0,0 +1,3 @@
+# SPDX-FileCopyrightText: 2025 2025 wahl.chat
```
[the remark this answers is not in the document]
SimonKaran13: This should be 2026 instead of 2025 @copilot

ai-backend/src/ingestion/connector.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from .schemas import ChunkRecord
+
+
+class BaseConnector(ABC):
+    """Abstract base for all wahl.chat V2 data source connectors.
```
SimonKaran13: Let's not call it `wahl.chat V2`, just 'wahl.chat'

ai-backend/src/ingestion/connector.py
```diff
@@ -0,0 +1,149 @@
...
+        Raises:
+            ValueError: If the raw payload is malformed or produces no usable
+                        chunks (runner skip-and-continues; cursor does not advance).
+        """
+        ...
+
+    def post_upsert(
+        self, qdrant, collection_name: str, chunks: list[ChunkRecord]
+    ) -> int:  # noqa: ANN001
+        """Optional hook — called by the runner AFTER each successful item upsert.
```
SimonKaran13, line 132: Where is this hook actually called?

ai-backend/src/ingestion/ids.py
```diff
@@ -0,0 +1,42 @@
+# SPDX-FileCopyrightText: 2025 2025 wahl.chat
```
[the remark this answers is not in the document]
SimonKaran13: 2026

ai-backend/src/ingestion/governance_levels.py
```diff
@@ -0,0 +1,34 @@
...
+framework→connector layering). The framework (``retrieve.py``) imports from
+here; connectors may re-export from this module (the abgeordnetenwatch taxonomy
+keeps its own identical definitions — a parity test pins the equality).
+
+Semantics copied EXACTLY from
+``connectors/abgeordnetenwatch/topic_taxonomy_config.py``:
+``ALL_LEVELS`` is the {federal, state, municipal} relevance fallback used when
+a vote's topic cannot be classified — ``EU`` is deliberately NOT part of it
+(AW carries no EU parliament data; adding it would change down-rank behavior).
+"""
```
SimonKaran13, line 17: Let's remove this and just name the file `governance_levels.py`

ai-backend/src/ingestion/party_slugs.py
```diff
@@ -0,0 +1,87 @@
...
+  (plenary speeches only ever carry Bundestag fractions, so the speech connector
+  deliberately does NOT include these — an unexpected state party in a speech
+  label quarantines rather than silently mis-tenants).
+
+The **CDU/CSU rule** (do NOT "simplify"): a joint ``"cdu/csu"`` fraction label
+maps to ``"cdu"`` (the Union fraction, not splittable from the source), while a
+standalone ``"csu"`` label maps to ``"csu"`` (a distinct tenant). Collapsing CSU
+into CDU would mis-tenant CSU speeches/manifestos and break cross-source
+alignment.
+"""
```
SimonKaran13, line 15: Let's reduce the ammount of comments a bit. For this kind of files I think the code is self descriptive

ai-backend/src/ingestion/registry.py
```diff
@@ -0,0 +1,107 @@
...
+        OpenParliamentTvConnector,
+    )  # noqa: PLC0415
+
+    return OpenParliamentTvConnector()
+
+
+# ---------------------------------------------------------------------------
+# CONNECTOR_FACTORIES — the closed registry (no dynamic module path).
+# ---------------------------------------------------------------------------
+
```
SimonKaran13: What is the advantage of using this dict instead of using the connectors' constructors directly?

cc @copilot

ai-backend/src/ingestion/run.py
```diff
@@ -0,0 +1,803 @@
...
+        )
+    return RunReport(
+        processed=processed,
+        remaining=remaining,
+        chunks_upserted=chunks_upserted,
+        failed_ids=tuple(failed_ids),
+        present_skips=present_skips,
+        chunks_unbekannt=chunks_unbekannt,
+    )
+
```
SimonKaran13, line 693: Let's refactor this function into smaller functions. Currently it has 3 nested for-loops which makes it very hard to read. Prefer documenting the code with more, smaller and meaningfully named functions instead of writing walls of comments.

(Add this to your agents.md 🤗)

ai-backend/src/ingestion/run.py
```diff
@@ -0,0 +1,803 @@
...
+# ---------------------------------------------------------------------------
+# __main__ — CLI dispatch on CONNECTOR_ID / --connector arg
+# ---------------------------------------------------------------------------
+
+
+if __name__ == "__main__":
+    # -----------------------------------------------------------------------
+    # Local-dev convenience: load ai-backend/.env if present so OPENAI_API_KEY
+    # (and any other local secrets) are available for the embed step. This is
```
SimonKaran13: Is the whole main function a local-dev convenience or just the hint on loading the .env?

SimonKaran13 (changes requested):
Let's move the ingestion code to it's own module in the repo. Currently it is inside of /ai-backend but it has no code dependencies to it. To keep the repo cleaner, let's have it organized as follows:

. (root)
|- web
|- ai-backend
|- ingestion
|-...

ai-backend/src/ingestion/__init__.py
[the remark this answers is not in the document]
SimonKaran13: On all new files not only here @copilot

SimonKaran13 (changes requested):
I reviewed the exact `stack/01-tooling` → `stack/02-corpus-foundation` diff at `199f9bd`, including every existing thread, and found five net-new blockers: the actual uv migration lands here without a runnable/testable container or backend; Qdrant 1.18 breaks the legacy retrieval API; the local Gemini configuration is loaded after the model constants are frozen; the collection lifecycle can silently mix OpenAI and Gemini vectors; and an empty normalized parent leaves stale chunks retrievable.

I also left two non-blocking architecture/tooling comments about tenant-selective retrieval and the stale lockfile.

Validation: `uv lock --check`, `uv sync --frozen`, Ruff, and 99 corpus-focused tests passed against live Qdrant. The full suite fails during collection (`socketio` missing), mypy fails on three removed `QdrantClient.search()` calls, documented startup has no `src.app` and the legacy startup lacks `aiohttp`, and the Docker build fails while copying the missing `poetry.lock`. The four green checks are Vercel-only and do not cover these backend paths.

I intentionally did not repeat the existing comments about copyright years, naming, comments, registry indirection, `run.py` structure/local-dev scope, connector hooks, file naming, module placement, or Gemini Embedding 2 task formatting.

ai-backend/pyproject.toml
```diff
@@ -1,46 +1,55 @@
-# SPDX-FileCopyrightText: 2025 2025 wahl.chat
+# SPDX-FileCopyrightText: 2026 wahl.chat
 #
 # SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
 
-[tool.poetry]
+[project]
```
SimonKaran13, line 5: [P1] The actual Poetry → uv conversion lands in this PR, but this exact stacked head no longer has a runnable backend. `uv sync --frozen` succeeds, then the full pytest run dies during collection because `socketio` was removed while the tests and runtime still import it; `uv run python -m src.app` fails because `src/app.py` is not in this stack item; and the Dockerfile still copies `poetry.lock` and runs Poetry. Please move the complete packaging/runtime migration into #114 (the tooling baseline), or keep this PR compatible with its base until that migration is complete.

ai-backend/pyproject.toml
```diff
@@ -1,46 +1,55 @@
...
+    "langchain-openai>=1.2.0",
+    "langchain-qdrant>=1.1.0",
+    "openai>=2.41.0",
+    "pydantic>=2.10.0,<3.0",
+    "pypdf>=6.0.0,<7.0",
+    "python-dotenv>=1.0.1,<2.0",
+    "pyyaml>=6.0.2",
+    "xxhash>=3.5.0",
+    # qdrant-client tracks Qdrant server minor 1.18.x (pin held for index parity)
+    "qdrant-client>=1.18.0,<1.19",
```
SimonKaran13, line 33: [P1] Pinning `qdrant-client` to 1.18 breaks the existing retrieval path: the frozen environment reports `hasattr(QdrantClient, "search") == False`, while `src/vector_store_helper.py` still calls `.search()` at lines 171, 311, and 343. `uv run mypy src` fails on all three calls, and a legacy request would raise at runtime. Please migrate those calls to the supported query API (or retain a compatible client) and add a regression that exercises the real client surface.

ai-backend/src/ingestion/run.py
```diff
@@ -0,0 +1,803 @@
...
+    # conditional on the file existing — Cloud Run Jobs have no .env, so env
+    # there continues to come from the job spec (do NOT use utils.load_env(),
+    # which hard-requires API_NAME and would crash the Cloud Run path).
+    # override=False keeps any explicitly-exported env (e.g. CONNECTOR_ID) authoritative.
+    from pathlib import Path  # noqa: PLC0415
+    from dotenv import load_dotenv  # noqa: PLC0415
+
+    _env_path = Path(__file__).resolve().parents[2] / ".env"
+    if _env_path.exists():
+        load_dotenv(_env_path, override=False)
```
SimonKaran13: [P1] This loads `.env` after the module has already imported `get_embeddings` and frozen `EMBEDDING_MODEL`/`EMBEDDING_DIM` from `setup_collection`. On the exact head, loading `.env.example` yields provider `gemini` and model `gemini-embedding-2`, but `get_embeddings()` constructs the Gemini client with the already-bound model `text-embedding-3-large`. Local ingestion therefore sends an OpenAI model name to Gemini. Load dotenv before configuration is imported, or resolve model/dimension lazily, and cover the real CLI startup order.

ai-backend/src/ingestion/setup_collection.py
```diff
@@ -0,0 +1,250 @@
...
+                      creation — should never occur on a healthy Qdrant).
+    """
+    if client is None:
+        client = _make_client()
+    # -----------------------------------------------------------------------
+    # Existence guard — mirrors vector_store_helper.py lines 148-155.
+    # -----------------------------------------------------------------------
+    existing = [c.name for c in client.get_collections().collections]
+    if COLLECTION_NAME in existing:
+        print(f"Collection '{COLLECTION_NAME}' already exists — skipping creation.")
```
SimonKaran13, line 344: [P1] Reusing the same `wahlchat_chunks_<env>` name is unsafe for the OpenAI → Gemini migration. An existing 3072-dimensional OpenAI collection reaches this branch, skips creation, and has no stored model/provider fingerprint to reject Gemini writes or queries. Because Gemini is also configured at 3072 dimensions, the dimension guard passes and the collection can silently contain two incompatible vector spaces. Use a fresh versioned collection plus full re-ingest/alias cutover, or persist and verify the complete embedding-space fingerprint before any write or query.

ai-backend/src/ingestion/run.py
```diff
@@ -0,0 +1,803 @@
...
+                                    key="source_item_id",
+                                    match=models.MatchAny(any=superseded_siids),
+                                )
+                            ]
+                        )
+                    ),
+                    wait=True,
+                )
+
+            if chunks:
```
SimonKaran13: [P1] `BaseConnector.normalize()` explicitly allows an empty list, but all parent-footprint reconciliation is behind this condition. If a parent previously produced child A and later normalizes to `[]`, the runner performs no footprint scroll and no delete, so A remains retrievable forever. I reproduced this with an empty connector (`delete_called=False`, zero footprint scrolls). Treat an authoritative empty result as an empty parent footprint (while keeping failures distinguishable) and add an A → empty regression test.

ai-backend/src/ingestion/retrieve.py
```diff
@@ -0,0 +1,882 @@
...
+                    corpus without an HNSW index — forbidden.
+    """
+    # Reject non-selective searches BEFORE embedding or touching the client.
+    # The collection's HNSW is configured m=0 / payload_m=16 (tenant-only, per
+    # CLAUDE.md "What NOT to Do"): there is NO global HNSW index, so an
+    # unfiltered vector search degrades to a full O(n) brute-force scan.
+    # Only party_id, party_ids_contains, and source_type count as selective —
+    # region_path / authority_tier / publish_after alone are NOT
+    # sufficient to avoid the global scan.
+    _is_selective = (
```
SimonKaran13, line 447: [P2] This guard does not enforce the tenant filter the collection design relies on. `party_id` is the only `is_tenant` field, but `source_type` or `party_ids_contains` alone passes; the vote path is even documented to use `party_ids_contains` instead of `party_id`. With global HNSW disabled (`m=0`), those no-tenant searches take the global/group-scan path the guard claims to forbid. Add the actual tenant condition for every query (for example a shared vote tenant), or change the index design and guard so the accepted filters match the available HNSW paths.

ai-backend/uv.lock
```diff
@@ -0,0 +1,1814 @@
+version = 1
```
SimonKaran13, line 1: [P2] This lock is not current despite the repository instruction to use the latest dependencies. Running `uv lock --upgrade` against the live index updates direct packages including FastAPI 0.136.3 → 0.140.0, LangChain 1.3.4 → 1.3.14, `langchain-google-genai` 4.2.4 → 4.3.1, `langchain-openai` 1.2.2 → 1.4.1, and OpenAI 2.41.0 → 2.48.0. Please regenerate the lock with the current index and rerun the compatibility checks.

ai-backend/src/ingestion/run.py
```diff
@@ -0,0 +1,803 @@
...
+        collection_name: Target Qdrant collection.
+        chunks:          ChunkRecord instances to upsert.
+        vectors:         Corresponding embedding vectors (one per chunk).
+
+    Raises:
+        DimensionMismatchError: If any vector's dimension != EMBEDDING_DIM.
+    """
+    from qdrant_client.models import PointStruct  # noqa: PLC0415
+
+    points = []
```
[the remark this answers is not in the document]
annikasiefke: Resolved with `zip(..., strict=True)`

ai-backend/src/embeddings.py
annikasiefke: langchain-google-genai forwards `task_type` to the embed API. We stick to the docs guidance one choosing RETRIEVAL_QUERY vs RETRIEVAL_DOCUMENT (runner embeds with RETRIEVAL_DOCUMENT, retrieval path with RETRIEVAL_QUERY)

ai-backend/src/ingestion/connector.py
annikasiefke: `run_connector()` calls `post_upsert()` after the upsert batch loop completes (src/ingestion/run.py), the openparliament connector uses it for the DIP supersede pass

ai-backend/pyproject.toml
annikasiefke: Accepted as is for now.

ai-backend/pyproject.toml
annikasiefke: Accepted as is for now (`vector_store_helper.py` is V1 code that is deleted by stack/07; the breakage exists only on intermediate heads. )

ai-backend/src/ingestion/run.py
annikasiefke: the `main()` entrypoint is the production entrypoint (Cloud Run Job should run `python -m src.ingestion.run --connector …`); only the `.env` loading is local-dev convenience

SimonKaran13 (approved):

SimonKaran13 (approved):

PR #116, opened 2026-07-25
V2 stack 3/9: Abgeordnetenwatch votes connector

annikasiefke:
V2 stack 3/9 — the first real connector: Abgeordnetenwatch roll-call votes, the canonical `vote_record` producer.

> **Diff-size note:** ~18.5k of the +22.7k lines are one captured AW API response (`fixtures/aw_poll_3602_votes.json` — the full per-MdB vote list for poll 3602 that the mapper tests replay). It's marked `linguist-generated`, so it renders collapsed. The **actual reviewable diff is ~4.2k lines across 17 files**, and half of that is tests.

- **AW API client** with the 2s/request rate limit and paged poll discovery.
- **Vote mapping**: one chunk per poll with per-party tallies and stances; `relevance_levels` tagging from the topic taxonomy so federal votes can be surfaced (down-ranked) in state chats where the topic actually applies.
- **`LEGISLATURE_CONFIG`**: federal WP19–21 plus all 16 Landtage, current + prior period each, with AW period IDs cited per row.

Two AW test files (`test_corpus_chunk_poll.py`, `test_legislature_config.py`) arrive in the manifestos PR instead of here — they assert region mapping through the manifesto mapper's `region_for_period`, so they'd break the build at this point in the stack.

Verified: ruff, mypy, 179 tests, GDPR wall guard.

### DIFF
--- a/ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
+++ b/ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2025 wahl.chat
+# SPDX-FileCopyrightText: 2026 wahl.chat
 #
 # SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
 
@@ -10,8 +10,10 @@ This connector implements the 3-method synchronous ABC
 normalize() — no Firestore, no GCS, no work_queue, no matcher dual-write.
 
 The runner (run.py) owns embed + upsert; this connector is pure data-transform.
-The cursor (since) is derived by the runner from Qdrant max(external_id) for
-"vote_record" and passed to discover() — no watermark methods needed.
+The runner still passes its Qdrant-derived cursor to discover() (ABC contract),
+but discover() deliberately IGNORES it: discovery is a per-legislature
+set-difference against the store (see discover()), which re-surfaces failures
+a max(external_id) watermark would permanently skip.
 
 Security notes:
     GDPR Art. 9 wall: NO code path reads users/{uid}.  This invariant
@@ -93,7 +95,10 @@ class AbgeordnetenwatchVotesConnector(BaseConnector):
                         None (default) reads AW_LEGISLATURE_ID from the
                         environment at CONSTRUCTION time (not import time, so a
                         long-lived process can construct one connector per
-                        legislature), falling back to 111 (19th Bundestag).
+                        legislature). REQUIRED: when neither the argument nor
+                        the env var is set, construction fails — a silent
+                        default would quietly limit a production run to one
+                        historical period while newer periods never ingest.
     """
 
     # Class attribute: used by runner.get_cursor() to scope the Qdrant scroll.
@@ -107,7 +112,17 @@ class AbgeordnetenwatchVotesConnector(BaseConnector):
         # would freeze the value for the process lifetime and force one process
         # invocation per legislature.
         if legislature_id is None:
-            legislature_id = int(os.getenv("AW_LEGISLATURE_ID", "111"))
+            raw_env = os.getenv("AW_LEGISLATURE_ID", "").strip()
+            if not raw_env:
+                raise ValueError(
+                    "AbgeordnetenwatchVotesConnector needs an explicit "
+                    "legislature: pass legislature_id or set AW_LEGISLATURE_ID. "
+                    "There is deliberately no default — every configured period "
+                    "must be run explicitly (the make targets "
+                    "run-abgeordnetenwatch-votes / run-all-landtage-votes loop "
+                    "over all federal / Landtag periods)."
+                )
+            legislature_id = int(raw_env)
         self._legislature_id = legislature_id
 
         cfg = LEGISLATURE_CONFIG.get(legislature_id)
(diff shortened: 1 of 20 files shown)


ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/corpus.py
```diff
@@ -0,0 +1,377 @@
...
+        fraction_label: The ``label`` field from an AW ``fraction`` dict in
+                        a votes response, e.g. ``"FDP (Bundestag 2017 - 2021)"``.
+
+    Returns:
+        Canonical party slug string, e.g. ``"fdp"``, or ``"unbekannt"``.
+    """
+    # Strip parenthetical parliament-period suffix, then normalise away invisible
+    # formatting characters (U+00AD soft hyphen in the Greens' fraction name).
+    clean = _normalize_fraction_label(_PARENS_SUFFIX_RE.sub("", fraction_label))
+    return _AW_FRACTION_SLUG_MAP.get(clean, _PARTY_SLUG_QUARANTINE)
```
SimonKaran13: [P1] This map does not cover the Landtag periods configured in this PR. Live poll 6404 contains `FDP/DVP (Baden-Württemberg 2021 - 2026)`, and this function returns `unbekannt`; the live fraction catalogue also exposes `FDP (Gruppe)`, `FREIE WÄHLER (Gruppe)`, and `BVB - Freie Wähler` variants that miss the map. Those votes lose the real party from `party_ids`, and multiple misses are merged into one `unbekannt` tally. Add the known aliases and tests covering the configured Landtag labels, not only the old Bundestag fixture.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
```diff
@@ -0,0 +1,413 @@
...
+
+    def __init__(
+        self,
+        legislature_id: Optional[int] = None,
+    ) -> None:
+        # Env read lives in the __init__ BODY: an import-time default arg
+        # would freeze the value for the process lifetime and force one process
+        # invocation per legislature.
+        if legislature_id is None:
+            legislature_id = int(os.getenv("AW_LEGISLATURE_ID", "111"))
```
SimonKaran13: [P1] The silent fallback makes the documented/default federal run ingest only the 19th Bundestag. The live API currently has 162 polls for period 132 and 63 for the current period 161, but the registry/Make target constructs this connector without an ID and there is no corresponding all-Bundestag execution path. Require an explicit period or execute all configured federal periods; a production/default run must not quietly stop at historical period 111.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
```diff
@@ -0,0 +1,413 @@
...
+        This replaces a max(external_id) high-water mark, which permanently lost any
+        poll that was skipped/failed below the max on a prior run (the max advanced past
+        it and the strict ``id > cursor`` filter never re-surfaced it). Set-difference is
+        gap-free and self-healing: a poll missing from Qdrant is always re-attempted, and
+        a permanently-unprocessable poll simply stays absent without blocking newer polls.
+        """
+        from qdrant_client import models as qdrant_models
+        from src.ingestion.setup_collection import COLLECTION_NAME
+
+        qdrant = self._get_qdrant()
```
SimonKaran13, line 192: [P2] Discovery opens a second Qdrant client and hard-codes the module-level collection, even though `run_connector()` already receives the client and `collection_name` it will use for cursor reads and upserts. With an injected test client, a non-default collection, or a migration, discovery and writes can therefore hit different stores. Inject the runner's store/collection into this set-difference path (or move the store-side discovery logic into the runner) so one run has one Qdrant contract.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/connector.py
```diff
@@ -0,0 +1,413 @@
...
+        # non-int id would crash str(p["id"]) sorting / fetch downstream.
+        polls = [p for p in polls if isinstance(p.get("id"), int)]
+
+        # Set-difference discovery: ingest polls present in the AW API but not yet in
+        # Qdrant for this legislature. Gap-free and self-healing — a poll skipped/failed
+        # on a prior run stays missing and is retried, and a permanently-unprocessable
+        # poll keeps failing harmlessly without blocking newer polls (no high-water mark
+        # to corrupt).
+        #
+        # AW_REFRESH=1 forces a full reconcile: return ALL polls so run.py's change-aware
```
SimonKaran13, line 303: [P2] `AW_REFRESH=1` refreshes parents that still exist upstream, but it is not a full reconcile: a poll removed from the AW result never appears in `discover()`, so the runner never gets a parent key to delete and the old chunk remains retrievable indefinitely. Add an explicit source-vs-store prune/reconciliation step, or document that retractions are intentionally retained instead of calling this a full reconcile.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/corpus.py
```diff
@@ -0,0 +1,377 @@
...
+    relevance_levels = get_relevance_levels(
+        topic_labels, warn_on_missing_topics=(source_item.region == "DE")
+    )
+
+    # Stable hash of the correctness-bearing content (embed text + tallies/outcome
+    # + participating parties) PLUS the stamped envelope fields (region,
+    # publish_date, relevance_levels, wahlperiode, legislature_period_id) — so an
+    # AW_REFRESH reconcile run re-writes chunks whose envelope changed (e.g. a
+    # re-classified topic), not only text/tally corrections.
+    content_hash = hashlib.sha256(
```
SimonKaran13, line 373: [P2] `citation_url` is written to the chunk but omitted from the change detector. I reproduced that changing only `abgeordnetenwatch_url` produces the same `content_hash`; an `AW_REFRESH` run then treats the point as unchanged and preserves the stale citation. Include the citation URL in the canonical hash and cover that refresh case in a mapper/runner test.

SimonKaran13 (changes requested):
I am requesting changes because the connector currently produces incomplete or incorrectly attributed vote coverage on supported configurations.

Live Abgeordnetenwatch checks confirm that the row-count pagination implementation is correct: `range_start=100&range_end=200` returned 200 unique rows, and poll 3602 returned all 709 votes in one 800-row request. All 36 configured parliament-period IDs, labels, and dates match the live API, and the 53-topic taxonomy matches the live topic catalogue.

The blockers are in how those periods are used. A current configured Baden-Württemberg poll maps `FDP/DVP` to `unbekannt`, with the same problem for several known Landtag fraction variants, so party-filtered retrieval loses real votes. Separately, the zero-configuration factory/Make path silently runs period 111 only; periods 132 and 161 are configured and live but have no federal execution path unless an operator knows to set an undocumented per-run ID.

I also left non-blocking comments about the connector bypassing the runner's Qdrant client/collection, refresh not pruning parents removed upstream, and citation URL changes being invisible to `content_hash`.

Validation on `91399ae`: 91 AW/enum tests passed; Ruff and mypy passed for the new connector files. The broader ingestion suite still fails during collection because `langchain_text_splitters` is not installed, which is inherited from #115 rather than introduced here. The four green PR checks are Vercel-only. #116 is still based on `8f0f1f9` while #115 is now `199f9bd`; a synthetic merge is clean, but the branch still needs restacking so GitHub reviews the declared stack state.

ai-backend/src/ingestion/connectors/abgeordnetenwatch/mappers/corpus.py
annikasiefke: Thanks, I missed that, is fixed now!

SimonKaran13 (approved):

SimonKaran13 (approved):

PR #117, opened 2026-07-25
V2 stack 4/9: Bundestag speeches (DIP) connector

annikasiefke:
V2 stack 4/9 — Bundestag plenary speeches from the DIP API (`parliamentary_speech`, `source="dip"`).

- **Protocol XML parser** for plenary transcripts, with speaker→party resolution via the official MdB master data (Stammdaten).
- **Speech chunking** (~1000-char boundary-aware) and PDF citations pointing at the original Bundestag transcript.
- **`speech_key` / `speech_dedup`** — the shared speech-identity infrastructure that the openparliament.tv connector (next-but-one PR) uses to dedupe against DIP.
- The runner's cross-connector tests land here too (`test_run.py` exercises the runner against this connector, including the disappeared-child parent-footprint reconciliation).

Scope note: speeches are deliberately capped at WP20/21 — earlier protocols have formats the parser can't ingest reliably — and there is no Landtag speech source, so this content is federal-only by nature.

This PR comes before manifestos (a swap vs. the originally proposed order) because the manifesto test suite contains the cross-connector slug-parity guard that asserts against this connector's party-slug map.

Verified: ruff, mypy, 330 tests, GDPR wall guard.

### DIFF
--- a/ai-backend/src/ingestion/connectors/bundestag_speeches/bulk.py
+++ b/ai-backend/src/ingestion/connectors/bundestag_speeches/bulk.py
@@ -1,0 +1,246 @@
+# SPDX-FileCopyrightText: 2025 wahl.chat
+#
+# SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
+
+"""
+Standalone bulk CLI for replaying Bundestag plenary speeches from a local
+JSONL dump into ``wahlchat_chunks_{ENV}`` via the relocated corpus mapper.
+
+This is the --from-jsonl backfill entrypoint that keeps the existing 99 MB
+``speeches.jsonl`` replayable without re-fetching from the DIP API.  The
+live connector (``connector.py``) handles incremental runs via ``run.py``
+and the YYYYMMDD cursor; this script handles the one-shot historical backfill.
+
+Design notes:
+  - This is the ONLY file within the ``bundestag_speeches`` package permitted
+    to call the runner ``_embed_texts`` / ``_upsert_chunks`` helpers directly.
+    The connector never touches those helpers — ``run.py`` owns that path.
+  - Replayed rows pre-date the live connector, so ``build_chunk_records``
+    leaves ``external_id=None`` for the bulk path (the live connector stamps
+    YYYYMMDD in ``normalize()``; a future migration patches legacy rows).
+  - The script is idempotent: deterministic chunk UUIDs (via ``compute_chunk_id``)
+    mean re-running upserts the same Qdrant point IDs, which is a no-op.
+
+Usage (local dev):
+    cd ai-backend
+    uv run python -m src.ingestion.connectors.bundestag_speeches.bulk --help
+    uv run python -m src.ingestion.connectors.bundestag_speeches.bulk --limit 20
+    uv run python -m src.ingestion.connectors.bundestag_speeches.bulk speeches.jsonl
+    QDRANT_URL=http://localhost:6333 ENV=dev \\
+        uv run python -m src.ingestion.connectors.bundestag_speeches.bulk
+
+JSONL path resolution (CLI arg > SPEECHES_JSONL env > repo-root default):
+    1. Positional ``jsonl_path`` argument (also accepted as ``--from-jsonl``).
+    2. ``SPEECHES_JSONL`` environment variable.
+    3. Repo root ``speeches.jsonl`` (three levels above ``ai-backend/``).
+
+Requires: make stores-up first (Qdrant running at QDRANT_URL).
+          OPENAI_API_KEY in ai-backend/.env (auto-loaded if present).
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import os
+import sys
+from pathlib import Path
+from typing import Optional
+
+from langchain_openai import OpenAIEmbeddings
+from qdrant_client import QdrantClient
+
+from datetime import date as _date_type
+
+from src.ingestion.connectors.bundestag_speeches.mappers.corpus import (
+    build_chunk_records,
+)
+from src.ingestion.run import _embed_texts, _upsert_chunks
+from src.ingestion.schemas import ChunkRecord
+from src.ingestion.setup_collection import COLLECTION_NAME, EMBEDDING_MODEL
+
+logger = logging.getLogger(__name__)
+
+_BATCH_SIZE = 100  # chunks per embed+upsert batch
+
+
+# =============================================================================
+# MAIN INGEST LOOP
+# =============================================================================
+
+
+def ingest(
+    jsonl_path: Path,
+    qdrant: QdrantClient,
+    embed: OpenAIEmbeddings,
+    limit: Optional[int] = None,
+) -> tuple[int, int]:
+    """Read speeches from *jsonl_path* and upsert into Qdrant in batches.
+
+    Each JSONL line is a speech dict passed through
+    ``build_chunk_records`` from the relocated corpus mapper.  Replayed
+    rows are stamped with ``external_id=YYYYMMDD`` so that
+    ``get_cursor("parliamentary_speech")`` picks up the backfill and the
+    first live incremental run does not re-embed everything.
+
+    Args:
+        jsonl_path:  Path to the speeches JSONL file (one JSON object per line).
+        qdrant:      Initialised QdrantClient.
+        embed:       Initialised OpenAIEmbeddings instance.
+        limit:       Maximum number of speeches to process (None = all).
+
+    Returns:
+        Tuple of (speeches_processed, chunks_upserted).
+    """
+    speeches_processed = 0
+    chunks_total = 0
+    batch: list[ChunkRecord] = []
+
+    def _flush(b: list[ChunkRecord]) -> int:
+        if not b:
+            return 0
+        vectors = _embed_texts(embed, [c.text for c in b])
+        _upsert_chunks(qdrant, COLLECTION_NAME, b, vectors)
+        return len(b)
+
+    with jsonl_path.open(encoding="utf-8") as fh:
+        for raw_line in fh:
+            raw_line = raw_line.strip()
+            if not raw_line:
+                continue
+
+            if limit is not None and speeches_processed >= limit:
+                break
+
+            speech = json.loads(raw_line)
+            records = build_chunk_records(speech)
+            if not records:
+                # Skip empty-text or unparseable-date speeches (no chunk emitted, not counted).
+                continue
+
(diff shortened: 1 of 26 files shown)


ai-backend/src/ingestion/connectors/bundestag_speeches/connector.py
```diff
@@ -0,0 +1,704 @@
...
+        protocol: dict = raw["protocol"]
+        speeches: list[dict] = raw.get("speeches", [])
+        mdb_lookup: dict = raw.get("mdb_lookup") or {"by_id": {}, "by_name": {}}
+
+        # Build speech rows, dropping non-MdB speakers (999…) and empty-text speeches
+        rows: list[dict] = []
+        for speech in speeches:
+            if not speech.get("text"):
+                continue
+            if _is_non_mdb_speaker(speech):
```
SimonKaran13: [P1] This removes real government speeches, not just procedural chair remarks. The Bundestag DTD says `9999…` IDs cover all non-MdB speakers, including federal ministers, state secretaries, state ministers, and Bundesland ministers. I reproduced it with the official [20/214 XML](https://www.bundestag.de/resource/blob/1057624/20214.xml): `ID2021402000` is Alexander Schweitzer as Ministerpräsident of Rheinland-Pfalz, the parser extracts 11,337 characters, and this branch drops the whole speech. Preserve these records and resolve attribution through an explicit role/person mapping (or a reviewed unknown-party policy) instead of classifying the entire ID range as non-content; the official definition is on pp. 13–14 of the [Bundestag DTD documentation](https://www.bundestag.de/resource/blob/577234/dbtplenarprotokoll_kommentiert.pdf).

ai-backend/src/ingestion/connectors/bundestag_speeches/connector.py
```diff
@@ -0,0 +1,704 @@
...
+
+    # The DIP cursor must NOT be scoped to source="dip". op progressively
+    # supersede-DELETES dip points, so max(external_id, source="dip") walks
+    # BACKWARD as coverage grows (worst case None → the run re-fetches and
+    # re-parses the whole Wahlperiode every time). op external_ids are the same
+    # YYYYMMDD scale and op only ever supersedes a dip twin that WAS ingested,
+    # so the cross-source max over parliamentary_speech is a valid,
+    # non-regressing floor; the lookback window still re-covers recent gaps.
+    # (op keeps its own source-scoped cursor via the BaseConnector default.)
+    cursor_source = None
```
SimonKaran13, line 377: [P1] The cross-source maximum is not a valid initial or partial DIP-backfill cursor. I reproduced the exact connector with a stored OP/WP21 cursor of `20260710`: selecting DIP WP20 sends `f.datum.start=2026-05-11` and discovers zero WP20 protocols. The same happens when OP is populated before DIP, so the stated WP20/21 coverage becomes order-dependent and requires an operator to know about `--since`. Keep a durable DIP + Wahlperiode coverage watermark outside deletable points (so OP superseding DIP cannot move it backward), rather than borrowing the newest date from another source/period.

ai-backend/src/ingestion/connectors/bundestag_speeches/connector.py
```diff
@@ -0,0 +1,704 @@
...
+                    continue
+                docs.append(doc)
+
+        # Client-side floor filter: DIP server-side f.datum.start is an optimisation;
+        # post-filter here for correctness since the fake client (and early DIP responses)
+        # may return protocols before the floor. Uses the SAME lookback-adjusted floor
+        # as the server-side param so a failed-below-max protocol is re-discovered.
+        if floor_int is not None:
+            floor_iso = _yyyymmdd_int_to_iso(floor_int)
+            docs = [d for d in docs if (d.get("datum") or "") >= floor_iso]
```
SimonKaran13: [P1] A bounded event-date lookback does not make failed protocols or corrections durable. `run_connector()` continues after a failure, so later successes advance `max(external_id)`; once the failed protocol is more than 60 days behind that max it is filtered out forever. This is reachable with the default batch size: the latest 30 official WP21 protocols span 10 July back to 4 March 2026. The same filter also never revisits an old protocol corrected today, although DIP exposes `f.aktualisiert.start` specifically for update polling. Track failures durably and use an update watermark/full reconciliation path; sorting oldest-first alone does not prevent the cursor from advancing past a skip.

ai-backend/src/ingestion/connectors/bundestag_speeches/bulk.py
```diff
@@ -0,0 +1,246 @@
...
+        # ai-backend/src/ingestion/connectors/bundestag_speeches/bulk.py → parents[5]
+        resolved_path = Path(__file__).resolve().parents[5] / "speeches.jsonl"
+
+    if not resolved_path.exists():
+        print(f"ERROR: JSONL file not found: {resolved_path}", file=sys.stderr)
+        sys.exit(1)
+
+    qdrant_url = os.getenv("QDRANT_URL", "http://localhost:6333")
+    _qdrant = QdrantClient(url=qdrant_url)
+    _embed = OpenAIEmbeddings(model=EMBEDDING_MODEL)
```
SimonKaran13: [P1] The documented backfill bypasses the configured embedding contract. `.env.example` selects Gemini and the generic runner calls `get_embeddings(task_type="RETRIEVAL_DOCUMENT")`, but this module imports `EMBEDDING_MODEL`/`COLLECTION_NAME` before loading `.env` and then always constructs `OpenAIEmbeddings`. In the documented local flow that either requires an unrelated OpenAI key or writes OpenAI vectors while the corpus/query path expects Gemini, corrupting retrieval. Load configuration before importing frozen constants and use the same provider factory and task type as `run.py`.

ai-backend/src/ingestion/connectors/bundestag_speeches/bulk.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    speeches_processed = 0
+    chunks_total = 0
+    batch: list[ChunkRecord] = []
+
+    def _flush(b: list[ChunkRecord]) -> int:
+        if not b:
+            return 0
+        vectors = _embed_texts(embed, [c.text for c in b])
+        _upsert_chunks(qdrant, COLLECTION_NAME, b, vectors)
```
SimonKaran13: [P2] This replay is only point-ID idempotent; it is not replacement/reconciliation safe. It neither stamps `source_parent_key` nor compares/deletes the prior footprint, so a corrected JSONL speech that shrinks leaves higher-index chunks behind, and a speech removed from the dump remains retrievable indefinitely. Because this is the explicit historical-backfill path, route it through the runner's parent-footprint logic or implement an equivalent source-vs-store prune before calling it idempotent.

ai-backend/src/ingestion/connectors/bundestag_speeches/mappers/corpus.py
```diff
@@ -0,0 +1,300 @@
...
+            file=sys.stderr,
+        )
+        return []
+
+    party_slug = party_to_slug(speech.get("party"))
+
+    speaker_name = speech.get("speaker_name") or ""
+    protocol_id = speech.get("protocol_id") or ""
+    citation_title = f"{speaker_name}, {raw_date} (Protokoll {protocol_id})"
+    citation_url: Optional[str] = speech.get("pdf_url") or None
```
SimonKaran13, line 226: [P2] The official XML already provides precise citation coordinates: each table-of-contents `xref ref-type="rede"` maps the speech's `xml_rede_id` to printed page `pnr` and quadrant `div`. In the official 21/90 and 20/214 samples, every parsed `rede` ID had that mapping, but the parser discards it and this mapper links only the bare protocol PDF, so a citation opens at the document start and does not identify the supporting page. Thread page/quadrant through the parser and metadata/citation title (and add a PDF deep link where the physical-page offset can be derived safely).

SimonKaran13 (changes requested):
I reviewed the exact declared `stack/03-votes` → `stack/04-dip-speeches` diff at `8860990` (one commit, 23 new files) and am requesting changes for four authored correctness blockers.

The live connector can skip the DIP history entirely when openparliament.tv or a newer Wahlperiode populated the shared speech cursor first; the 60-day event-date watermark can also permanently lose older failed protocols and later transcript corrections. The speaker filter drops real government speeches because Bundestag `999…` IDs include non-MdB ministers and state secretaries, not just procedural officials. Finally, the JSONL backfill ignores the configured Gemini embedding path and constructs OpenAI directly after importing frozen configuration.

I also left non-blocking comments about the bulk path bypassing parent-footprint reconciliation and the parser discarding page/quadrant data that the official XML exposes for every tested speech. Separately, `mdb.py` directly imports `httpx` and `tenacity`, but they are not declared as production dependencies (`httpx` is dev-only and `tenacity` is currently only transitive).

Reproducible validation on the exact head: 151 DIP/speech-key/resurrection/runner tests passed; Ruff passed; mypy passed on all 12 new source files; `uv lock --check` passed. The full suite still fails during collection because `socketio` is absent, inherited from #115. Against official live data, the parser extracted 112/112 speeches from 21/90 with zero unresolved parties after loading the current 15.3 MB MdB master file; 20/214 reproduced the dropped 11,337-character minister-president speech. The current DIP OpenAPI confirms cursor pagination, `f.wahlperiode`, `f.datum.start`, `f.aktualisiert.start`, and `fundstelle.xml_url/pdf_url`.

#117's immediate base SHA still exactly matches #116's head, so the 23-file authored diff is clean and I found no immediate-base restack noise. It is nevertheless transitively stale: #116 is still based on `8f0f1f9` while #115 now points at `199f9bd`; restack #116 first, then #117.

I intentionally did not repeat the three existing Copilot threads, #115's unresolved empty-normalize/Qdrant/tenant/vector-space blockers, or #116's existing connector-store and citation-hash comments.

ai-backend/src/ingestion/connectors/bundestag_speeches/connector.py
annikasiefke: Thanks, fixed it!

ai-backend/src/ingestion/connectors/bundestag_speeches/connector.py
annikasiefke: I fixed this for now in the same way we handle the votes, i.e. to avoid a watermark stored outside we're doing a set-difference against the store.

ai-backend/src/ingestion/connectors/bundestag_speeches/mappers/corpus.py
annikasiefke: Nice, thanks. Is added now!

SimonKaran13 (approved):

PR #118, opened 2026-07-25
V2 stack 5/9: Party manifestos connector

annikasiefke:
V2 stack 5/9 — party manifestos (`party_manifesto`) from the abgeordnetenwatch election-program API.

- **HTML scrape / PDF parse** with page metadata, so citations can deep-link to the cited page; citations always point at the original publisher — we never re-host files.
- **Chunking via the shared boundary-aware chunker**; `region` tagged per election (federal programs `DE`, Landtagswahl programs `DE-<state>`), which is what makes level-scoped retrieval work downstream.
- **Cross-connector slug-parity guard**: the same real party must resolve to the same `party_id` across votes, speeches, and manifestos — otherwise tenant-filtered retrieval silently returns nothing. This test is why manifestos come after speeches in the stack.
- Includes the two AW test files deferred from the votes PR (they assert region mapping via this connector's `region_for_period`).

The connector has a local dev CLI (`bulk.py`, `--dry-run`/`--ids`/`--limit`) alongside the standard registry path.

Verified: ruff, mypy, 490 tests, GDPR wall guard.

### DIFF
--- a/ai-backend/src/ingestion/connectors/manifestos/bulk.py
+++ b/ai-backend/src/ingestion/connectors/manifestos/bulk.py
@@ -1,0 +1,381 @@
+# SPDX-FileCopyrightText: 2025 wahl.chat
+#
+# SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
+
+"""
+Standalone bulk CLI for ingesting party manifestos (Wahlprogramme) from the
+Abgeordnetenwatch ``election-program`` API into ``wahlchat_chunks_{ENV}``.
+
+This is the bespoke local-dev entrypoint that the generic runner (run.py) cannot
+express: it adds the --dry-run / --ids / --limit flags on top of the same data
+path as ManifestoConnector.  The connector's interface methods are driven by
+run.py for the standard cursor-based embed path; this script reuses the
+connector's shared I/O helpers (determine_source, load_program_pages,
+_fetch_period_date) and the pure mappers so there is a single fetch/parse +
+record-build implementation.
+
+Source rule (locked design):
+  1. If ``link[0].uri`` is non-null  -> scrape HTML main content (trafilatura).
+  2. Elif ``file`` (PDF url) non-null -> download + parse the PDF in-memory.
+  3. Else skip the program.
+
+citation_url points at the original source URL in both cases — wahl.chat does
+not re-serve or store the files (no local storage, no Firestore source-doc).
+
+Scope: by default ALL programs are ingested. Pass ``--since YYYY-MM-DD`` (or set
+``MANIFESTO_SINCE``) to floor by parliament_period ``election_date``; the operator
+chooses the window and the CLI carries no built-in cut-off.
+
+Usage (local dev):
+    cd ai-backend
+    uv run python -m src.ingestion.connectors.manifestos.bulk --dry-run --limit 5
+    uv run python -m src.ingestion.connectors.manifestos.bulk --ids 598,599,600
+    uv run python -m src.ingestion.connectors.manifestos.bulk --since 2020-01-01
+    QDRANT_URL=http://localhost:6333 ENV=dev uv run python -m src.ingestion.connectors.manifestos.bulk
+
+Re-running REWRITES each processed program's footprint: the flush deletes the
+program's existing chunks by source_item_id (wait=True) before upserting the
+fresh ones, so a shrunk/replaced PDF leaves no stale higher-index chunks behind.
+Deterministic chunk UUIDs (via compute_chunk_id) keep unchanged chunks at the
+same point IDs; everything flushed IS re-embedded (this bespoke CLI has no
+content-hash skip — use run.py + MANIFESTO_REFRESH for the cheap reconcile).
+"""
+
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+import sys
+from datetime import date as date_type
+from typing import Optional
+
+from langchain_core.embeddings import Embeddings
+from qdrant_client import QdrantClient
+
+from src.ingestion.connectors.manifestos.connector import (
+    _fetch_period_date,
+    determine_source,
+    load_program_pages,
+    resolve_since_floor,
+)
+from src.ingestion.connectors.manifestos.mappers.corpus import (
+    SourceKind,
+    build_manifesto_records,
+    chunk_pages,
+    party_to_slug,
+)
+from src.embeddings import get_embeddings
+from src.ingestion.run import _embed_texts, _upsert_chunks
+from src.ingestion.schemas import ChunkRecord
+from src.ingestion.setup_collection import COLLECTION_NAME
+
+logger = logging.getLogger(__name__)
+
+_BATCH_SIZE = 100  # chunks per embed+upsert batch
+
+
+# =============================================================================
+# MAIN INGEST LOOP
+# =============================================================================
+
+
+def ingest(
+    qdrant: QdrantClient,
+    embed: Embeddings,
+    limit: Optional[int] = None,
+    ids: Optional[list[int]] = None,
+    dry_run: bool = False,
+    since: Optional[date_type] = None,
+) -> tuple[int, int]:
+    """Fetch manifesto programs from AW, embed, and upsert into Qdrant.
+
+    Args:
+        qdrant:  Initialised QdrantClient.
+        embed:   Initialised embeddings client (from get_embeddings()).
+        limit:   Max programs to process (None = all passing the date filter).
+        ids:     Specific AW program IDs to process (overrides limit).
+        dry_run: If True, skip embed/upsert; fetch, parse, and print per-program
+                 stats (party, period, source_kind, pages, chars, chunks).
+        since:   Optional election-date floor; programs whose parliament_period
+                 election_date is strictly before it are skipped. When None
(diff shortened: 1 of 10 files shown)


ai-backend/src/ingestion/connectors/manifestos/connector.py
```diff
@@ -0,0 +1,510 @@
...
+    links = program.get("link") or []
+    link_uri: Optional[str] = None
+    if links and isinstance(links, list):
+        first = links[0]
+        if isinstance(first, dict):
+            link_uri = first.get("uri") or None
+
+    file_url: Optional[str] = program.get("file") or None
+
+    if link_uri:
```
SimonKaran13: [P1] Selecting a single source before fetching makes the advertised HTML/PDF fallback impossible. The live API currently has 112 programs with both a usable link and a file; for program 129 I reproduced that the selected archived HTML viewer resolves to HTTP 404 while the PDF returns HTTP 200. `load_program_pages()` raises on the HTML response and `fetch()` records a skip without ever trying the file, so the default all-history run silently loses a document that is available. Return ordered source candidates and fall back to the PDF on fetch/extraction failure, with a regression using a failing link plus a successful file.

ai-backend/src/ingestion/connectors/manifestos/connector.py
```diff
@@ -0,0 +1,510 @@
...
+            )
+            if not date_iso:
+                continue
+            try:
+                election_date = date_type.fromisoformat(date_iso)
+            except (ValueError, TypeError):
+                continue
+            if since_floor is not None and election_date < since_floor:
+                continue
+            if pid in ingested_ids:
```
SimonKaran13, line 430: [P1] One stored chunk is not proof that the whole program committed. `_upsert_chunks()` writes in 256-point slices, and live Bundestag program 600 normalizes to 317 chunks. If the first slice succeeds and the second fails, the next run collects external ID 600 here and excludes it, so the missing tail is never repaired without an operator knowing to set `MANIFESTO_REFRESH`. Track a completed-parent marker, or let discovered parents reach the runner's full point-ID/footprint comparison, and cover a first-slice-success/second-slice-failure retry.

ai-backend/src/ingestion/connectors/manifestos/bulk.py
```diff
@@ -0,0 +1,381 @@
...
+        # (no delete-before-embed loss window; mirrors run.py's ordering).
+        vectors = _embed_texts(embed, [c.text for c in b])
+
+        # Rewrite semantics (mirrors run.py's footprint guard) — delete each
+        # program's EXISTING footprint by source_item_id before upserting, one
+        # delete per program with wait=True, so a re-run against a shrunk PDF
+        # (fewer chunks than stored) leaves no stale higher-index chunks
+        # retrievable forever.
+        siids = sorted({str(c.source_item_id) for c in b})
+        for siid in siids:
```
SimonKaran13: [P1] Embedding first does not close the loss window because the entire old footprint is still deleted before any replacement upsert succeeds. I reproduced `delete → upsert(exception)`; the existing program is absent afterward. Program 600 also spans two Qdrant upsert requests, so the first new slice can commit and the second fail after the old footprint is gone. Upsert deterministic surviving/new point IDs first, then delete only the old orphan IDs after all writes succeed, or route this path through the runner's footprint reconciliation. Add failure injection at both the first and later upsert request.

ai-backend/src/ingestion/connectors/manifestos/bulk.py
```diff
@@ -0,0 +1,381 @@
...
+    # Load ai-backend/.env if present.
+    from pathlib import Path  # noqa: PLC0415
+
+    from dotenv import load_dotenv  # noqa: PLC0415
+
+    _env_path = Path(__file__).resolve().parents[4] / ".env"
+    if _env_path.exists():
+        # .env is the source of truth for local dev — let it win over any stale
+        # value inherited from the shell.
+        load_dotenv(_env_path, override=True)
```
SimonKaran13: [P1] This new entrypoint repeats the frozen-configuration failure from the shared runner and also breaks explicit store selection. `get_embeddings`, `run`, and `setup_collection` were imported before this call; reproducing the real order with the example config gives provider `gemini` but frozen model `text-embedding-3-large` and frozen collection `wahlchat_chunks_dev`. `override=True` can then replace a `make run-manifestos QDRANT_URL=… ENV=prod` shell target with the local `.env` URL, while line 355 omits `QDRANT_API_KEY` entirely. Load dotenv before importing configuration, preserve explicit environment values, and construct Qdrant/embeddings through the same factory as the registry path; test the actual module startup order for local and remote settings.

ai-backend/src/ingestion/connectors/manifestos/mappers/corpus.py
```diff
@@ -0,0 +1,446 @@
...
+    party_slug = party_to_slug(party_label)
+    region = region_for_period(period_label)
+    wahlperiode = wahlperiode_for_period(period_label)
+
+    publish_date = date_type.fromisoformat(period_date_iso)
+
+    sid = compute_source_item_id("party_manifesto", str(program_id))
+
+    # citation_url is the original source URL (external PDF or weblink).
+    citation_url = source_url
```
SimonKaran13: [P2] The extracted page never reaches the citation link. Every PDF record keeps the same bare document URL, and no runtime consumer reads `meta.page_start` to add an anchor; my built record had page 1 in metadata but no `#page=` in `citation_url`. The PR's page-aware citation claim therefore still opens every manifesto at the first page. Build a chunk-specific PDF URL (preserving any existing query/fragment safely) or make the citation transport explicitly render `page_start`, and assert the final URL emitted to the client.

ai-backend/src/ingestion/connectors/manifestos/mappers/corpus.py
```diff
@@ -0,0 +1,446 @@
...
+            raw_party_label=party_label if party_slug == "unbekannt" else None,
+        ).model_dump(mode="json", exclude_none=True)
+
+        # Change-aware content_hash (mirrors the op mapper): per-chunk text so an
+        # updated program text re-writes via run.py's guard; includes the resolved
+        # party slug (indexed tenant field). NOTE: on the connector path the
+        # guard only sees already-present programs on MANIFESTO_REFRESH=1 runs
+        # (normal discover excludes them via set-difference); bulk.py rewrites
+        # each program's footprint unconditionally.
+        content_hash = hashlib.sha256(
```
SimonKaran13, line 433: [P2] A refresh cannot correct provenance-only changes because the change detector omits the fields it is meant to refresh. I reproduced identical hashes after changing the source URL, page 1→7, and total page count while keeping text/party/region fixed; the runner then treats the point as unchanged and preserves the stale citation/meta. Include the displayed citation URL/title and meaningful source/page/publish metadata in the canonical hash, with a `MANIFESTO_REFRESH` regression where text is unchanged but provenance changes.

SimonKaran13 (changes requested):
I reviewed the exact declared `stack/04-dip-speeches` → `stack/05-manifestos` diff at `cc3c581` (one commit, 10 new files) and am requesting changes for four authored correctness blockers.

The source selection drops a usable PDF whenever the preferred HTML URL is stale; this is already present in the live catalogue. Set-difference discovery treats one stored chunk as a completed program, so a partial multi-request upsert becomes permanent. The standalone bulk CLI freezes embedding/collection configuration before loading `.env` and can override the store explicitly selected by the operator. Finally, that same CLI deletes the prior parent footprint before the replacement upsert succeeds, so a Qdrant failure causes corpus loss.

I also left non-blocking comments because page numbers never become PDF deep links and the manifesto `content_hash` ignores displayed provenance. The bulk module duplicates runner-owned embedding/upsert/reconciliation behavior; routing non-dry-run writes through the shared runner, or extracting one shared commit primitive, would remove most of these divergence points.

Live validation against the official Abgeordnetenwatch API found 635 election programs across 46 periods: 501 PDF-only, 22 link-only, and 112 with both a usable link and file. Program 129's archived HTML viewer currently resolves to HTTP 404 while its PDF returns 200. Current program 653 parsed as 29 pages / 94 chunks; current HTML program 651 produced 2,597 characters / 4 chunks; Bundestag program 600 parsed as 72 pages / 317 chunks, reproducing the multi-upsert partial-commit path.

On the exact head, 160 manifesto plus deferred AW tests passed; Ruff passed on all five new source files; mypy passed on all five; `uv lock --check` and the GDPR wall passed. The full suite still fails during collection because `socketio` is absent, inherited from #115. GitHub exposes only the successful Vercel checks for this head; no backend/frontend GitHub Actions run is attached.

#118's base SHA still exactly matches #117's head, so the authored diff is clean with no immediate-base restack noise. It remains transitively stale: #116 is based on `8f0f1f9` while #115 now points at `199f9bd`; restack #116, then #117, then #118.

I intentionally did not repeat #115's inherited packaging/Qdrant/empty-normalize/tenant/vector-space findings, #116's existing live fraction-alias/default-period/store-client/full-source-reconcile findings, or #117's speaker/cursor/watermark/JSONL/citation/dependency findings. The provenance-hash note below is on #118's separately authored manifesto mapper, not the inherited #116 implementation.

SimonKaran13 (approved):

PR #119, opened 2026-07-25
V2 stack 6/9: openparliament.tv speech enrichment

annikasiefke:
V2 stack 6/9 — openparliament.tv speech enrichment (`parliamentary_speech`, `source="op"`).

- **Keyless bulk connector** over openparliament.tv's published data — the same Bundestag speeches as DIP, but with sentence-level timing and a video deep-link per speech.
- **Cross-source supersede**: duplicates are detected via the shared `speech_key`; the richer op.tv version supersedes its DIP twin, grafting the DIP transcript PDF onto it first so nothing is lost. Speech `source_item_id`s carry the connector source (`dip`/`op`), so the two id spaces structurally can't collide.
- **`src/deeplink.py`** — the sentence-level video-timestamp locator the chat backend uses to seek the plenary video to the cited moment.

Verified: ruff, mypy, 525 tests, GDPR wall guard.

### DIFF
--- a/ai-backend/src/ingestion/connectors/openparliament_tv/bulk.py
+++ b/ai-backend/src/ingestion/connectors/openparliament_tv/bulk.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2025 wahl.chat
+# SPDX-FileCopyrightText: 2026 wahl.chat
 #
 # SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
 
@@ -30,8 +30,9 @@ Usage (local dev):
     QDRANT_URL=http://localhost:6333 ENV=dev \\
         uv run python -m src.ingestion.connectors.openparliament_tv.bulk --from-github
 
-Requires: make stores-up first (Qdrant running at QDRANT_URL).
-          OPENAI_API_KEY in ai-backend/.env (auto-loaded if present).
+Requires: make stores-up first (Qdrant running at QDRANT_URL) and the
+          configured embedding provider's API key in ai-backend/.env
+          (auto-loaded; exported shell env wins).
 """
 
 from __future__ import annotations
@@ -43,12 +44,24 @@ import sys
 from pathlib import Path
 from typing import Optional
 
-from langchain_openai import OpenAIEmbeddings
+# CLI startup ONLY: load ai-backend/.env BEFORE the imports below — embeddings
+# and setup_collection freeze EMBEDDING_MODEL / EMBEDDING_DIM / COLLECTION_NAME
+# from the environment at import time, so a later load_dotenv() configures
+# nothing. override=False keeps explicitly-exported shell env authoritative.
+if __name__ == "__main__":
+    from dotenv import load_dotenv
+
+    _bulk_env_path = Path(__file__).resolve().parents[4] / ".env"
+    if _bulk_env_path.exists():
+        load_dotenv(_bulk_env_path, override=False)
+
+from langchain_core.embeddings import Embeddings
 from qdrant_client import QdrantClient
 
+from src.embeddings import get_embeddings
 from src.ingestion.connector import BaseConnector
 from src.ingestion.connectors.bundestag_speeches.constants import MDB_STAMMDATEN_FILE
-from src.ingestion.connectors.bundestag_speeches.mdb import load_mdb_lookup
+from src.ingestion.connectors.bundestag_speeches.mdb import ensure_mdb_lookup
 from src.ingestion.connectors.openparliament_tv.client import OpTvClient
 from src.ingestion.connectors.openparliament_tv.mappers.corpus import (
     build_chunk_records,
@@ -56,9 +69,10 @@ from src.ingestion.connectors.openparliament_tv.mappers.corpus import (
 from src.ingestion.connectors.openparliament_tv.supersede import (
     supersede_dip_duplicates,
 )
+from src.ingestion.ids import compute_chunk_id
 from src.ingestion.run import _embed_texts, _upsert_chunks
 from src.ingestion.schemas import ChunkRecord
-from src.ingestion.setup_collection import COLLECTION_NAME, EMBEDDING_MODEL
+from src.ingestion.setup_collection import COLLECTION_NAME, check_fingerprint
 
 # Shared speech-dedup helper — one definition for both speech connectors.
 from src.ingestion.speech_dedup import datum_to_external_id as _datum_to_external_id
@@ -110,7 +124,7 @@ _MDB_PATH: Path = Path(__file__).resolve().parents[4] / MDB_STAMMDATEN_FILE
 def ingest(
     client: OpTvClient,
     qdrant: QdrantClient,
-    embed: OpenAIEmbeddings,
+    embed: Embeddings,
     mdb_lookup: dict,
     limit: Optional[int] = None,
 ) -> tuple[int, int]:
@@ -123,7 +137,7 @@ def ingest(
     Args:
         client:     Initialised OpTvClient.
         qdrant:     Initialised QdrantClient.
-        embed:      Initialised OpenAIEmbeddings instance.
+        embed:      Embeddings client from ``get_embeddings()``.
         mdb_lookup: MdB-Stammdaten lookup for the empty-faction / CSU fallback.
         limit:      Maximum number of usable speeches to process (None = all).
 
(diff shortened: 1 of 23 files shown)


ai-backend/src/ingestion/connectors/openparliament_tv/connector.py
```diff
@@ -0,0 +1,288 @@
...
+                continue
+
+            items = [it for it in (session.get("data") or []) if isinstance(it, dict)]
+            ext = _session_external_id(items)
+            if ext is None:
+                continue
+            if floor is not None and ext < floor:
+                # Descending scan: every remaining basename is older still —
+                # stop fetching entirely.
+                break
```
SimonKaran13: [P1] This early stop assumes reverse lexical filename order is chronological, but the live tree disproves that invariant. It starts with the special files `21902`, `21901`, `21900`, `21800`; `21902` is dated 2026-02-23, while the later filename `21090` is dated 2026-07-10. On this exact head, `discover(20260710)` fetched only `21902-session.json`, compared it to the 2026-05-11 floor, hit this `break`, and returned `[]`, so it never reached current sessions 21090/21089/etc. Derive and sort the actual session dates before applying an early stop, or keep scanning without using filename order as a date proof. Please cover a special old filename followed by a newer normal session in the discovery regression.

ai-backend/src/ingestion/connectors/openparliament_tv/supersede.py
```diff
@@ -0,0 +1,245 @@
...
+        op_norm = _norm_speech_text(_join_indexed_texts(op["texts"]))
+        if not op_norm:
+            continue
+        for dsid, dip in dip_speeches.items():
+            if dip["key"] != op["key"] or dsid in delete_dip_sids:
+                continue
+            dip_norm = _norm_speech_text(_join_indexed_texts(dip["texts"]))
+            if not dip_norm:
+                continue
+            ratio = difflib.SequenceMatcher(None, op_norm, dip_norm).ratio()
```
SimonKaran13: [P1] The default `SequenceMatcher` enables `autojunk`, which treats frequent characters as junk on these long alphanumeric strings and makes the 0.85 policy fail on real twins. Comparing the official [OP 21/90 session](https://raw.githubusercontent.com/OpenParliamentTV/OpenParliamentTV-Data-DE/main/processed/21090-session.json) with the official [Bundestag 21/90 XML](https://dserver.bundestag.de/btp/21/21090.xml), only 15 of 76 same-key pairs clear 0.85 here (median 0.572); the same pairs with auto-junk disabled are all 0.922–0.961. The other 61 twins remain duplicated, and the inverse DIP resurrection guard currently uses the same comparator. Use a similarity method suitable for long normalized speeches and keep both directions symmetric; add an official-length near-identical regression rather than only short fixture strings.

ai-backend/src/ingestion/connectors/openparliament_tv/bulk.py
```diff
@@ -0,0 +1,267 @@
...
+    if not args.from_github:
+        print(
+            "ERROR: nothing to do — pass --from-github to run the GitHub bulk backfill.",
+            file=sys.stderr,
+        )
+        sys.exit(2)
+
+    qdrant_url = os.getenv("QDRANT_URL", "http://localhost:6333")
+    _qdrant = QdrantClient(url=qdrant_url)
+    _embed = OpenAIEmbeddings(model=EMBEDDING_MODEL)
```
SimonKaran13: [P1] This backfill does not use the corpus's configured embedding/store contract. The module imports `COLLECTION_NAME` and `EMBEDDING_MODEL` before loading `.env`, then always constructs `OpenAIEmbeddings`; meanwhile the documented config selects Gemini and the registry path calls `get_embeddings(task_type="RETRIEVAL_DOCUMENT")`. The Qdrant client above also omits `QDRANT_API_KEY`. In the documented flow this can write OpenAI vectors into a collection queried as Gemini (or the frozen dev collection), require an unrelated OpenAI key, and fail against authenticated dev/prod Qdrant. Load configuration before resolving frozen settings, construct embeddings through the shared factory with the same task type, and pass the API key. Please exercise the real module startup order for local and remote configs.

ai-backend/src/ingestion/connectors/openparliament_tv/bulk.py
```diff
@@ -0,0 +1,267 @@
...
+    _client = OpTvClient()
+    _mdb_lookup = load_mdb_lookup(_MDB_PATH)
```
SimonKaran13: [P1] A clean checkout does not contain the gitignored `data/MDB_STAMMDATEN.XML`, and `load_mdb_lookup` only warns and returns empty maps. The live connector deliberately uses `ensure_mdb_lookup` because continuing would poison party attribution, but the historical backfill silently continues. I reproduced official 21/90 producing 10 of 107 usable chunks as `party_id="unbekannt"` with the clean-checkout path versus 0 of 107 after loading the current 15.3 MB master; those chunks disappear from party-tenant retrieval. Fetch/validate the master here too (or require an explicit valid path) and fail before writing when the lookup is empty. Cover the fresh-checkout backfill path.

ai-backend/src/ingestion/connectors/openparliament_tv/bulk.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+    speeches_processed = 0
+    chunks_total = 0
+    batch: list[ChunkRecord] = []
+
+    def _flush(b: list[ChunkRecord]) -> int:
+        if not b:
+            return 0
+        vectors = _embed_texts(embed, [c.text for c in b])
+        _upsert_chunks(qdrant, COLLECTION_NAME, b, vectors)
```
SimonKaran13, line 157: [P2] This is point-ID idempotency, not source replacement. The backfill neither stamps `source_parent_key` nor compares the previous session footprint, so rerunning after an upstream speech disappears or a long speech shrinks only overwrites the current IDs; removed speeches and higher-index chunks stay retrievable indefinitely. Route writes through the runner's commit/reconcile path, or implement the same replacement-safe session footprint: preserve existing points until the replacement upserts succeed, then delete only confirmed orphans. The expected regressions are session A+B → A and one speech shrinking from two chunks to one.

ai-backend/src/ingestion/connectors/openparliament_tv/mappers/corpus.py
```diff
@@ -0,0 +1,322 @@
...
+        transcript_type="proceedings",
+        aligned=True,
+        sentence_map=sentence_map or None,
+        creator=media.get("creator"),
+        license=media.get("license"),
+        source_data=_OP_SOURCE_DATA,
+    ).model_dump(mode="json", exclude_none=True)
+
+    # --- change-aware content_hash: re-alignment / correction re-writes the chunk ---
+    content_hash = hashlib.sha256(
```
SimonKaran13, line 277: [P2] The change detector omits fields that drive cross-source identity and the citation users see: `speech_key`, speaker name/Wikidata ID, agenda metadata, `source_page`, and the derived citation title. If OP corrects only a speaker or agenda label while the transcript/timing/video stay unchanged, the runner treats the point as unchanged and preserves the stale dedup key and citation metadata. Include the material emitted identity/provenance fields in the canonical hash and add a metadata-only refresh regression that proves the stored `speech_key` and citation are replaced.

SimonKaran13 (changes requested):
I reviewed the exact declared `stack/05-manifestos` → `stack/06-openparliament` diff at `bece0a3` (one commit, 22 new files) and am requesting changes for four authored correctness blockers.

The incremental connector currently stops before reaching recent normal sessions because special session filenames are not chronological. The supersede comparator also misses most real OP/DIP twins on official 21/90 data because `SequenceMatcher`'s default auto-junk heuristic is unsuitable for long normalized character strings. The standalone backfill independently repeats the frozen/hard-coded embedding path and omits remote Qdrant auth. On a clean checkout it also silently runs without the gitignored MdB master, which mis-tenants real chair/minister speeches as `unbekannt`.

I also left non-blocking comments because the backfill bypasses parent-footprint reconciliation and the mapper's change detector omits cross-source identity/citation metadata.

Reproducible validation on the exact head: all 50 OP/supersede/deeplink/runner-focused tests passed; Ruff passed on the new source and test files; mypy passed on all 11 new source files; and the GDPR wall passed. The full backend suite still fails during collection because `socketio` is undeclared, inherited from #115. The live OP tree currently returns 3,162 entries / 318 matching session files with `truncated=false`. Official 21/90 data produced 107 usable OP speeches; its direct MP4 links return HTTP 200 with byte ranges, and stored sentence timestamps are relative to those speech videos. GitHub exposes only the four successful Vercel checks for this head.

#119's base SHA exactly matches #118's current head, so this 3,510-line authored diff has no immediate-base restack noise. The stack remains transitively stale because #116 still targets obsolete #115 SHA `8f0f1f9` instead of current `199f9bd`; restack #116 and descendants before merge.

I intentionally did not repeat Copilot's malformed-`sentence_map` thread, #115's inherited empty-normalize/tenant/vector-space findings, #117's shared DIP cursor/lookback/speaker findings, or #118's manifesto bulk/citation findings. The bulk comments below are on #119's separately authored OP entrypoint.

SimonKaran13 (approved):

PR #120, opened 2026-07-25
V2 stack 7/9: Transport migration — aiohttp/Socket.IO to FastAPI + SSE chat backend

annikasiefke:
V2 stack 7/9 — the transport migration: the V1 aiohttp/Socket.IO app (`aiohttp_app.py`, `websocket_app.py`) is replaced by a FastAPI backend with SSE chat, grounded in the corpus built by PRs 2–6.

- **Socket.IO → SSE**: `POST /api/v1/chat` streams Vercel AI SDK v5 UI-message-stream parts over plain HTTP (SSE works through corporate networks that block WebSockets). FastAPI replaces aiohttp: typed routes, Pydantic validation, bounded request DTOs.
- **Chat RAG pipeline** (`chat_service` / `chatbot_async`): per-party retrieval fan-out across votes/manifestos/speeches, two-pass current+historic temporal split, region/level scoping — manifestos are level-exclusive (a state chat cites only its own Landtag program), federal votes appear down-ranked in state chats with an explicit disclosure instruction.
- **Routes**: chat, voting-behavior (real `vote_record` tallies), pro/con, misc; Firebase auth; heartbeats and graceful SSE error frames.
- **`REQUIRE_CORPUS` rollout gate**: opt-in fail-fast — a deployment that sets it refuses to boot on a missing/empty corpus, so chat can never silently serve an empty store. Unset locally/CI.
- Dockerfile + entrypoint for Cloud Run.

Ordering note (deviation from the proposed stack, where transport was step 2): the V2 chat service imports the corpus retrieval, the votes legislature config, and the speech deep-link module at module level — a transport-first PR couldn't build without throwaway stub versions of those modules that later PRs would delete. Serving after sources needs zero shim code; only the review order changes.

Verified: ruff, mypy, 577 tests + 10-test in-process SSE smoke suite, GDPR wall guard.


ai-backend/src/routes/chat.py
```diff
@@ -0,0 +1,122 @@
...
+    "x-vercel-ai-ui-message-stream": "v1",
+    "Cache-Control": "no-cache, no-transform",
+    "X-Accel-Buffering": "no",
+}
+
+
+# Server-side bounds on the PUBLIC, unauthenticated chat endpoint. Every field
+# flows into LLM prompts, so an unbounded body is a cost / context-overflow /
+# memory vector. V1 enforced a 500-char user message (ChatUserMessageDto) that
+# the SSE migration dropped; restore it and bound the list inputs + history
```
romman8: prefer timeless comments

ai-backend/src/chat_service.py
```diff
@@ -111,202 +393,565 @@ async def emit_cached_party_response(
...
+    except Exception as _err:  # noqa: BLE001
+        logger.warning(
+            "%s() failed (source=%s party=%s): %s",
+            _log_prefix,
+            kwargs.get("source_type"),
+            kwargs.get("party_id") or kwargs.get("party_ids_contains"),
+            _err,
+            exc_info=True,
+        )
+        return []
```
SimonKaran13, line 491: [P1] Do not turn a corpus outage into a successful ungrounded answer

This catches every retrieval exception and makes it indistinguishable from a legitimate empty result. I forced all three source calls to raise: the stream still emitted sources_ready with an empty list, invoked the LLM, and finished with party_complete.status.indicator == "success". Keep per-source degradation, but track success/failure separately and fail the turn when every required retrieval failed. The SSE sequence should contain a generic terminal error and an error party_complete for the selected responder, followed by finish-step, finish, and [DONE]; it must not invoke or cache a "successful" answer. Add a failure-injection regression for that exact sequence.

ai-backend/src/chat_service.py
```diff
@@ -22,83 +51,336 @@
...
+def _mk_speech_docs(payloads: list[dict], improved_rag_query: str) -> list[Document]:
+    # op speeches carry a video deep-link (meta.sentence_map + video_uri);
+    # rewrite their url to video_uri#t={ts_start}. DIP unchanged.
+    return [
+        Document(
+            page_content=(p.get("text") or ""),
+            metadata={
+                "document_name": p.get("citation_title"),
+                "document_publish_date": p.get("publish_date"),
+                "url": _speech_deeplink_url(p, improved_rag_query),
```
SimonKaran13, line 209: [P1] Treat persisted sentence_map metadata as untrusted at this boundary

#120 is the first stack item to call the deeplink helper while assembling a live chat response. A Qdrant payload with sentence_map set to a string or containing a non-object entry raises AttributeError here and turns the whole party stream into an error. I reproduced that with sentence_map="corrupt". Validate/coerce the metadata before refinement and fall back to the stored citation_url or video_uri when it is malformed. The regression should still emit sources_ready, the normal party response, and [DONE], with the unrefined source URL rather than an error response.

ai-backend/src/chat_service.py
```diff
@@ -111,202 +393,565 @@ async def emit_cached_party_response(
...
-                                "document_publish_date"
-                            ),
-                            "url": source_doc.metadata.get("url"),
-                            "source_document": source_doc.metadata.get(
-                                "source_document"
-                            ),
-                            "party_id": rel_party.party_id,
-                        }
-                        sources.append(source)
+                        page_number = int(page_raw if page_raw is not None else 0) + 1
```
SimonKaran13: [P1] Build comparison citations through the same source-aware path

The new Document builders set pages to one-based values, so this increment changes page 1 to page 2. More importantly, this generic serializer drops the source-specific contract that the single-party branch now preserves: OP video links and attribution, the grafted transcript pdf_url, and the fields needed for post-answer deeplink refinement. I exercised this with a page-1 OP document carrying video and PDF URLs; the emitted source had page=2 and only the generic video URL. Use one typed citation builder for both branches, retain source/source_item_id and the OP/DIP-specific fields, and keep source indices stable if URLs are refined after generation.

ai-backend/src/chat_service.py
```diff
@@ -466,376 +1108,582 @@ async def process_party(
...
+                )
+                for p in parties_being_compared
+            ]
+            try:
+                # Parallel RAG doc fetch for comparison questions (NOT streaming output).
+                # The SSE stream output is SERIALIZED below.
+                await asyncio.wait_for(asyncio.gather(*party_tasks), timeout=40)
+            except asyncio.TimeoutError as e:
+                logger.error(f"Timeout fetching comparison docs: {e}")
+                yield _finish_step()
```
SimonKaran13, line 1661: [P1] Report timeouts as terminal errors before closing the stream

This emits a normal finish/DONE after responding_parties has already told the client to expect answers. The per-stream budget handler below does the same thing. #121 clears party loading on data-chat_event.error or an error party_complete; a normal finish leaves the responder pending until its 60-second watchdog. On both timeout paths emit a generic error event and, where a responder has started, party_complete with error status, then finish-step, finish, and [DONE]. Also assert that the timed-out task/generator is cancelled and that quick replies and cache writes do not run.

ai-backend/src/routes/pro_con.py
```diff
@@ -0,0 +1,97 @@
...
+_SSE_HEADERS = {
+    "Content-Type": "text/event-stream",
+    "Cache-Control": "no-cache, no-transform",
+    "X-Accel-Buffering": "no",
+    "x-vercel-ai-ui-message-stream": "v1",
+}
+
+
+@router.post("/pro-con")
+async def pro_con_endpoint(request: Request, body: ProConPerspectiveRequestDto):
```
SimonKaran13, line 40: [P1] Bound all inputs exposed by the new anonymous LLM routes

/chat adds useful bounds, but party_ids still has no per-item bound and this endpoint plus /voting-behavior accepts unbounded request IDs, party/context IDs, and last-user/assistant messages. The migrated JSON summary/swiper routes also accept unbounded histories. I confirmed both side-channel DTOs accept a one-million-character message before any provider call. Add field, item, depth, and aggregate limits and assert HTTP 422 for oversized inputs before retrieval or LLM work. A raw request-body limit at the proxy or ASGI edge is still needed because Pydantic only helps after the body has been parsed.

ai-backend/src/routes/voting_behavior.py
```diff
@@ -0,0 +1,319 @@
...
+                        continue
+
+                    complete_message += text_content
+
+                    # Split chunks per MAX_RESPONSE_CHUNK_LENGTH (preserves V1 chunking)
+                    for i in range(0, len(text_content), MAX_RESPONSE_CHUNK_LENGTH):
+                        if i > 0:
+                            await asyncio.sleep(0.025)
+                        split_chunk = text_content[i : i + MAX_RESPONSE_CHUNK_LENGTH]
+                        yield text_delta(body.request_id, split_chunk)
```
SimonKaran13, line 302: [P2] Either emit a valid v5 text block or document this as a custom SSE stream

The response advertises x-vercel-ai-ui-message-stream: v1, but these text-delta parts have no preceding text-start and no matching text-end. #121 currently hand-parses this endpoint, so it happens to work there, but a standard AI SDK v5 parser rejects a delta for an unopened text block. Emit text-start(request_id), then the deltas, then text-end(request_id), followed by voting_behavior_complete, finish-step, finish, and [DONE]; add an ordering assertion for that exact sequence.

ai-backend/src/utils.py
```diff
@@ -61,18 +68,32 @@ def safe_load_api_key(api_key: str) -> Optional[SecretStr]:
...
+def get_cors_allowed_origins(env: Optional[str]) -> list[str]:
+    """Enumerated CORS origins — NEVER "*".
+
+    app.py runs the CORS middleware with allow_credentials=True; a wildcard
+    there makes Starlette reflect ANY Origin with credentials, so origins must
+    always be an explicit list. CORS_EXTRA_ORIGINS (comma-separated) extends
+    the list per deployment (e.g. the Vercel URL of the deployed dev demo).
+    """
+    extra_origins = [
+        origin.strip()
```
SimonKaran13, line 87: [P2] Reject wildcard or invalid extra origins at startup

The new comment says this helper never returns "*", but CORS_EXTRA_ORIGINS=* puts it directly into allow_origins while allow_credentials=True. Starlette then treats the deployment as allow-all and mirrors credentialed preflights. Parse and validate http/https origins, explicitly reject "*", and fail startup on invalid entries so a production environment typo cannot silently broaden CORS.

SimonKaran13 (changes requested):
I reviewed the exact one-commit stack diff from bece0a336fec1a6a23428a96cacd89027f1fc55a to 876bde8c14705fee56b90e3721376c8ec11dceb2. The base is #119's current head, so these findings are on #120 rather than restack noise.

The FastAPI/SSE shape is coherent and the anonymous-versus-premium auth path is deliberate, but the failure and citation contracts are not safe to ship yet. A total Qdrant failure is reported as a successful answer, malformed persisted speech metadata can break response assembly, comparison citations are changed incorrectly, and timeouts look like normal completion to the frontend. The newly public LLM routes also remain partly unbounded. I left the concrete expected status/event behavior inline.

Validation on the exact head: uv sync --locked, Ruff, mypy, 598 pytest tests, the GDPR static wall, and the production Docker build all pass. The image boots and serves /healthz with explicit dummy provider configuration. I also exercised anonymous/auth gates, oversized DTOs, malformed sentence_map data, and forced retrieval failures. GitHub currently has only Vercel checks for this PR; none of the backend validation above runs as a #120 check.

ai-backend/src/auth.py
```diff
@@ -0,0 +1,95 @@
...
+import src.firebase_service  # noqa: F401
+
+logger = logging.getLogger(__name__)
+
+
+def verify_optional_bearer_token(request: Request) -> Optional[dict]:
+    """Verify an optional ``Authorization: Bearer <Firebase ID token>`` header.
+
+    Returns the decoded token claims when a valid token is present, else None.
+    NEVER raises and NEVER rejects the request — anonymous access stays
```
romman8, line 38: let's switch this asap in a follow-up to require anonymous firebase auth on each request

ai-backend/src/chatbot_async.py
```diff
@@ -391,6 +397,75 @@ def get_rag_context(relevant_docs: List[Document]) -> str:
...
+    """Convert vote_record payloads into LangChain Documents for numbered citation.
+
+    Each participating vote becomes one Document whose ``page_content`` is a German
+    block with the motion subject, outcome, source, and that party's tally + stance.
+    Votes where the party did not participate (``party_result`` is None) are skipped.
+
+    The resulting Documents flow into ``combined_docs`` and are numbered
+    sequentially by ``get_rag_context``, so the LLM can cite them as clean
+    ``[N]`` integer IDs instead of the free-text ``[Abstimmungsverhalten]``
+    marker that appeared when votes were injected as a raw heading block.
```
romman8: nit: omit the part about old faulty behavior from a previous internal iteration of the code

ai-backend/src/chatbot_async.py
```diff
@@ -446,7 +521,14 @@ async def generate_streaming_chatbot_response(
...
     context_id: str = DEFAULT_CONTEXT_ID,
     use_premium_llms: bool = False,
+    election_level: Optional[str] = None,
+    present_sources: Optional[tuple[bool, bool, bool]] = None,
+    has_historic: bool = False,
 ) -> AsyncIterator[BaseMessageChunk]:
+    # relevant_docs now contains V1 docs + manifesto_docs + speech_docs + vote_docs
+    # (combined_docs from chat_service.py); get_rag_context numbers them sequentially
+    # so the LLM can cite ALL of them as clean [N] integer IDs.  No free-text context
+    # kwargs remain — every grounding source flows through the numbered Document list.
```
romman8: nit: this also seems like it references a previous iteration of the code

ai-backend/src/routes/chat.py
```diff
@@ -0,0 +1,122 @@
...
+from src.chat_service import generate_chat_stream, with_heartbeat
+from src.models.chat import Role
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/api/v1")
+
+# SSE response headers
+# NOTE: Using StreamingResponse (not EventSourceResponse) because generate_chat_stream
+# yields pre-framed SSE strings ("data: ...\n\n"). EventSourceResponse would wrap them
```
romman8: What is the rationale for yielding pre-framed SSE strings?
I would prefer if we would reuse as much of FastAPI's out of the box SSE support as possible and use:
1. Their `ServerSentEvent` class https://fastapi.tiangolo.com/tutorial/server-sent-events/#serversentevent
2. EventSourceResponse

FastAPI also implements SSE best practices out of the box (EventSourceResponse) --> we should be able to drop the custom _SSE_HEADERS below and the custom `with_heartbeat`

ai-backend/tests/test_chat_sse.py
romman8, line 1: Would be awesome if you could add the following integration tests with mocked qdrant and LLM responses:
1. Single party response including citations with empty conversation history
2. Single party response including citations with non-empty conversation history
3. Multi-party responses with  non-empty conversation history
4. Pro-con perspective of a single party in the context of non-empty conversation history and a multiy-party response as the last turn
5. Comparison response

SimonKaran13 (approved):

PR #121, opened 2026-07-25
V2 stack 8/9: Next.js frontend

annikasiefke:
V2 stack 8/9 — the frontend's side of the transport migration: the Socket.IO client stack (`chat-socket.ts`, socket provider, connection-state actions, disconnected banner) is removed and the chat UI moves to SSE.

- **Chat UI on the AI SDK v5 `useChat` SSE transport** (the adapter for PR 7's stream format), with party selection, comparison mode, quick replies, and proposed questions.
- **Citation UI**: the sources dialog ("Quellen"), in-page **video viewer** seeked to the cited sentence, and **PDF viewer** deep-linked to the cited page behind a same-origin proxy with a host allowlist. This was sketched as its own stack step, but the root layout mounts the media-viewer provider and the chat renderer imports the sources dialog directly — a separate fixture-backed PR can't build without inventing an adapter the final code doesn't have, so it ships with the app.
- Voting-behavior detail view and pro/con perspectives carried over from V1, now backed by the ingested corpus.
- Supply-chain posture: `bun audit` gate clean at high severity (includes this week's `brace-expansion` OOM-DoS and `postcss` source-map path-traversal patches via overrides + reviewed release-age exemptions in `bunfig.toml`).

Verified: `next lint` + biome, `tsc --noEmit`, production build.


web/components/providers/sse-chat-provider.tsx
```diff
@@ -0,0 +1,191 @@
...
+    // v5: transport replaces the `api`/`streamProtocol` options. The backend
+    // emits the v5 UI-message-stream (data: <json> parts) so no protocol flag
+    // is needed — useChat parses it natively.
+    transport: new DefaultChatTransport({
+      api: '/api/chat',
+      // Attach the Firebase ID token per request (async resolvable): the proxy
+      // route forwards it so the backend can verify user_is_logged_in/premium
+      // Resolves to {} when signed out / on token errors —
+      // the request degrades gracefully to unauthenticated.
+      headers: () => getAuthHeader(),
```
SimonKaran13, line 86: [P1] Make exactly one layer own the Authorization header

AuthServiceWorkerProvider controls the same-origin `/api/chat` request, and `public/service-worker.js` already appends its Firebase token to every controlled request. This transport now sets Authorization first; the worker uses `Headers.append`, which serializes the result as `Bearer <page token>, Bearer <worker token>`. The backend passes that whole value to `verify_id_token`, verification fails, and a verified user is silently downgraded to the standard anonymous model. The worker also still signs into production Auth when the page SDK is in emulator mode. Remove one injection path or make the worker preserve/replace a single token and connect it to the selected Auth emulator. Cover a controlled-service-worker request where the backend receives exactly one valid token, plus the normal no-token case.

web/lib/firebase/firebase-emulators.ts
```diff
@@ -0,0 +1,18 @@
+// Local Firebase emulator wiring, shared by the browser and server client SDKs.
+//
+// Opt-in via NEXT_PUBLIC_USE_FIREBASE_EMULATORS=true (set in web/.env.local for
+// local development). When off, the SDKs talk to the real project exactly as
+// before — so production is unaffected and this is a no-op unless deliberately
+// enabled. Ports mirror firebase/firebase.json's emulators block.
+
+export const FIREBASE_EMULATOR_HOST = '127.0.0.1';
+export const FIRESTORE_EMULATOR_PORT = 8081;
+export const AUTH_EMULATOR_PORT = 9099;
```
SimonKaran13, line 10: [P1] Check in the emulator configuration these ports claim to mirror

`firebase/firebase.json` has no `emulators` block. I ran the documented `firebase emulators:start --project demo-wahl-chat --only firestore,auth` command: Firebase started Firestore on its default 8080 and printed that Auth was not configured, while `make stores-up` waits on 8081 and this module sends the browser to 8081/9099. Add the Firestore and Auth emulator host/port configuration and make the Makefile and example environment use the same source of truth. The local smoke test should start both processes, sign in anonymously, navigate between sessions, and confirm that no real Firebase project is contacted.

web/components/providers/sse-chat-provider.tsx
```diff
@@ -0,0 +1,191 @@
...
+      } else if (type === 'party_chunk') {
+        // V1: 'party_response_chunk_ready' event for secondary parties
+        mergeStreamingChunkPayloadForMessage(
+          ann.session_id as string,
+          ann.party_id as string,
+          ann as Parameters<typeof mergeStreamingChunkPayloadForMessage>[2],
+        );
+      } else if (type === 'party_complete') {
+        // V1: 'party_response_complete' event → completeStreamingMessage
+        completeStreamingMessage(
```
SimonKaran13: [P1] Branch on party_complete.status before persisting the answer

#120 emits this event with `status.indicator === "error"` for a per-party failure and does not need to emit the separate top-level `error` event. I fed that contract through the proxy and AI SDK parser; the status reaches this callback, but this branch ignores it, persists `complete_message` as an assistant answer, and counts the responder as successfully complete. On error, surface the failure and terminally clear or mark that announced responder without serializing the fallback as a normal answer. Cover a mixed two-party turn where one succeeds and one errors, and assert that neither party remains pending.

web/components/providers/sse-chat-provider.tsx
```diff
@@ -0,0 +1,191 @@
...
+          ann.sources as Parameters<typeof streamingMessageSourcesReady>[2],
+        );
+      } else if (type === 'party_chunk') {
+        // V1: 'party_response_chunk_ready' event for secondary parties
+        mergeStreamingChunkPayloadForMessage(
+          ann.session_id as string,
+          ann.party_id as string,
+          ann as Parameters<typeof mergeStreamingChunkPayloadForMessage>[2],
+        );
+      } else if (type === 'party_complete') {
```
SimonKaran13, line 143: [P1] Keep the turn busy until the stream is actually terminal

`party_complete` is party-scoped, not turn-terminal. When the last party reaches this branch, `completeStreamingMessage` clears `loading.newMessage`, but #120 can still spend time generating `quick_replies_title` before it emits finish. The input is therefore enabled during an active `useChat` request. AI SDK v5 does not serialize a second `sendMessage`; it replaces the active response, so trailing events from the first turn can be aborted or applied to the next one. Keep sends disabled until onFinish or another explicit terminal event, and reject concurrent sends. The regression should pause after the last party_complete, attempt a second send, and verify that the first title/quick replies stay attached to the first turn.

web/components/providers/sse-chat-provider.tsx
```diff
@@ -0,0 +1,191 @@
...
+    (state) => state.completeProConPerspective,
+  );
+  const resetStreamingMessageWatchdog = useChatStore(
+    (state) => state.resetStreamingMessageWatchdog,
+  );
+  const cancelStreamingMessages = useChatStore(
+    (state) => state.cancelStreamingMessages,
+  );
+
+  const { sendMessage, stop } = useChat({
```
SimonKaran13: [P2] Do not retain a second unscoped chat history inside useChat

The rendered and persisted history lives in Zustand and is already sent as `chat_history`, but DefaultChatTransport also serializes all of this hook's internal UI messages on every request. This provider survives navigation between chat sessions under the same context layout and the SDK messages are never keyed or reset by `chatSessionId`, so later requests include duplicate history from the previous session even though the proxy only needs the last user message. Scope/reset the hook per session or use `prepareSendMessagesRequest` to send only the current user message plus the explicit canonical DTO. Assert that a second turn and a same-context session switch each send one current SDK message and only the intended Zustand history.

web/lib/stores/actions/start-timeout-for-streaming-messages.ts
```diff
@@ -1,40 +1,79 @@
...
+// was not safely above their worst case.
 const STREAMING_MESSAGE_TIMEOUT =
-  Number(process.env.NEXT_PUBLIC_STREAMING_MESSAGE_TIMEOUT_MS) || 30000;
-const STREAMING_MESSAGE_CHECK_INTERVAL = 500;
-export const startTimeoutForStreamingMessages: ChatStoreActionHandlerFor<
-  'startTimeoutForStreamingMessages'
-> = (get, set) => async (streamingMessageId: string) => {
-  const clearIntervalAndTimeout = () => {
-    const { pendingStreamingMessageTimeoutHandler } = get();
+  Number(process.env.NEXT_PUBLIC_STREAMING_MESSAGE_TIMEOUT_MS) || 60000;
```
SimonKaran13, line 9: [P2] Align the configured watchdog with the backend's quiet periods

This fallback is 60 seconds because 30 seconds is documented here as unsafe, but the checked-in `.env.example` still sets `NEXT_PUBLIC_STREAMING_MESSAGE_TIMEOUT_MS=30000`, so normal local and copied deployment configuration never uses the new fallback. Backend heartbeat comments are consumed by the SSE parser and do not reach `onData`, while retrieval and quick-reply generation can be quiet within the 180-second budget. A valid slow response is therefore aborted at 30 seconds. Either reset from raw transport progress or choose a configured inactivity threshold consistent with the backend contract, and test a stream whose first data event and final quick-reply event each arrive after 30 seconds.

SimonKaran13 (changes requested):
I reviewed the exact one-commit stack diff from 876bde8c14705fee56b90e3721376c8ec11dceb2 to 1b1ee121e81e9b69b4456a1029be80a7537009d4. The base is #120's current head, so these findings are on #121 rather than restack noise.

The AI SDK v5 happy path reaches the browser in the expected event order, anonymous requests can omit Authorization, and the Socket.IO transport is gone. I am requesting changes because authenticated requests can carry two Bearer tokens, error party_complete events are persisted as successful answers, the UI accepts a second send before the first stream has finished, and the documented Firebase emulator stack does not start on the ports this frontend uses. The hidden AI SDK history and watchdog configuration also need to be bounded and scoped. Detailed expected behavior is inline.

Validation on the exact head: Bun's frozen install, lint, typecheck, production build, and high-severity audit pass; 10 backend SSE tests pass. The frontend package has no tests. A controlled backend fixture passed through the Next proxy and AI SDK parser in the expected start/data/text/finish order, including a parsed error party_complete. I also reproduced the Firebase CLI starting Firestore on 8080 and refusing to start Auth because firebase.json has no emulator configuration. The Vercel preview is deployment-protected, so source/dialog behavior could not be visually verified there.

SimonKaran13 (approved):

PR #122, opened 2026-07-25
V2 stack 9/9: Rollout scaffolding (draft)

annikasiefke:
V2 stack 9/9 — rollout scaffolding. **Draft on purpose**: the actual deployment workstream (Terraform for Cloud Run Jobs + Scheduler, secrets/Workload Identity, the one-time snapshot backfill into dev/prod Qdrant) doesn't exist yet — `infra/README.md` documents the plan. This PR firms up with that work; review PRs 1–8 first.

What's here:

- **`firebase/`**: Firestore security rules (chat-message privacy, tested against the emulator in CI), Cloud Functions, seed data + `seed_firestore.py` (hard-guarded: refuses to run without the emulator locally, refuses prod without an explicit confirm).
- **CI workflow** (backend / frontend / firebase-rules / smoke-test) — lands here because it's the first point in the stack where all four job targets exist.
- **`scripts/restore_qdrant_snapshot.sh`** — the tool for the initial corpus upload into dev/prod (snapshot create → upload → recover), with clobber guards.
- **München (Kommunalwahl 2026) context removed from the seed data** — abgeordnetenwatch has no municipal votes or speeches, so the context can't be meaningfully grounded yet; it returns with the manual-PDF-upload path for manifestos.

Merging this makes `develop` byte-identical to `v2-migration` (#111).


firebase/scripts/seed_firestore.py
```diff
@@ -32,72 +46,102 @@
...
+        )
+        sys.exit(1)
+else:
+    print(
+        f"ERROR: invalid SEED_TARGET={SEED_TARGET!r} — expected 'emulator' or 'real'."
+    )
+    sys.exit(1)
 
+# Configuration
+ENV = os.getenv("ENV", "dev")
```
SimonKaran13, line 102: [P1] Bind the fixture environment to the confirmed real project

With `SEED_TARGET=real GOOGLE_CLOUD_PROJECT=wahl-chat SEED_CONFIRM_PROJECT=wahl-chat` and no `ENV`, this defaults to `dev` and writes the development fixture set into production. The inverse mismatch is also accepted. For a real target, require `ENV` explicitly and validate a closed mapping such as `dev -> wahl-chat-dev` and `prod -> wahl-chat` before Firebase initialization; add subprocess tests asserting both mismatches exit before any client is created.

firebase/scripts/seed_firestore.py
```diff
@@ -272,7 +321,16 @@ def main():
     seed_parties(db)
 
     # Seed proposed questions for each context
-    seed_proposed_questions(db)
+    failed_files, failed_writes = seed_proposed_questions(db)
```
SimonKaran13, line 324: [P1] Make the München removal affect an existing Firestore project

This seed path only calls `.set()`; it never reconciles documents that disappeared from the fixture files. Deleting the München context, parties, and proposed-question files therefore leaves all previously seeded München documents live in dev/prod, even though the PR says the context is removed. Add an explicitly scoped removal migration or tombstone for that context, with a dry-run and the same project/environment confirmation, and prove against the emulator that unrelated V1 and V2 contexts remain untouched.

scripts/restore_qdrant_snapshot.sh
```diff
@@ -0,0 +1,85 @@
...
+[[ -n "$API_KEY" ]] && AUTH=(-H "api-key: ${API_KEY}")
+
+# Expand AUTH safely even when empty (bash 3.2 + set -u).
+q() { curl -sS ${AUTH[@]+"${AUTH[@]}"} "$@"; }
+
+echo "Target   : ${QDRANT_URL}  collection='${COLLECTION}'"
+echo "Snapshot : ${SNAPSHOT} ($(du -h "$SNAPSHOT" | cut -f1))"
+
+# Guard 1: existing non-empty collection needs --force.
+EXISTING=$(q "${QDRANT_URL}/collections/${COLLECTION}" || true)
```
SimonKaran13, line 56: [P1] Abort when the destructive preflight cannot inspect the target

`q` uses `curl -sS` without HTTP failure handling, this call suppresses its exit status, and the JSON parser maps every 401/403/5xx/network/malformed response to `MISSING`. The script can then reach the `priority=snapshot` upload without `--force` even though the existing collection was never safely inspected. Treat only an explicit Qdrant 404 as missing and fail closed for every other status or parse error. The regression should feed 401, 500, connection failure, and malformed JSON into the GET and assert that no POST is attempted.

scripts/restore_qdrant_snapshot.sh
```diff
@@ -0,0 +1,85 @@
...
+echo "Uploading + recovering ..."
+RESP=$(q -X POST \
+  "${QDRANT_URL}/collections/${COLLECTION}/snapshots/upload?priority=snapshot" \
+  -H "Content-Type:multipart/form-data" \
+  -F "snapshot=@${SNAPSHOT}")
+OK=$(printf '%s' "$RESP" | python3 -c \
+  "import sys,json;print((json.load(sys.stdin) or {}).get('status'))" 2>/dev/null || echo "")
+[[ "$OK" == "ok" ]] || die "recover failed: ${RESP}"
+
+# Verify.
```
SimonKaran13, line 81: [P1] Verify corpus compatibility and create a rollback point

A non-zero point count and green collection status do not establish that this is the intended corpus. A snapshot produced with another embedding model/provider can have the same 3072 width, and this check does not validate the named vector, distance, payload indexes, required source/legislature coverage, snapshot checksum, or target environment. `REQUIRE_CORPUS` also checks only count, so the wrong non-empty corpus would pass rollout. Require a versioned manifest/provenance check before recovery, snapshot the current target before any forced overwrite, verify schema plus expected source/period counts afterward, and print/test the rollback command.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,253 @@
...
+      - name: Bootstrap Qdrant collection
+        env:
+          QDRANT_URL: http://localhost:6333
+        run: uv run --no-sync python -m src.ingestion.setup_collection
+
+      - name: Unit / integration tests (excluding smoke test)
+        env:
+          QDRANT_URL: http://localhost:6333
+        run: |
+          uv run --no-sync pytest tests/ -x \
```
SimonKaran13, line 131: [P1] Run the seed tests instead of silently excluding/skipping them

The new production-target guard test is explicitly ignored here. `tests/test_seed.py` is included syntactically, but it module-skips unless Firestore is reachable on 8081 and this job never starts that emulator, so a green backend job proves neither the guard nor the fixture writes. Run `tests/test_local_mode.py` in this job and run `tests/test_seed.py` under `firebase emulators:exec --only firestore --project demo-wahl-chat` in a job with the locked Python environment; make an unexpected skip fail under CI.

infra/README.md
```diff
@@ -0,0 +1,109 @@
...
+  the 16 Landtage), the speech pair `bundestag_speeches,openparliament_tv` as a
+  single **sequential** job (they must never overlap), `manifestos` (infrequent /
+  on demand).
+- **Single image, `CONNECTOR_ID` selects the connector** (sequential for a
+  comma-separated list). The same container serves the FastAPI app (no
+  `CONNECTOR_ID`) and every ingestion job; dispatch is in
+  `ai-backend/docker-entrypoint.sh`.
+- **Task timeout 1h** (`--task-timeout 3600`) — the 15-minute cap is a Firebase
+  Scheduled Functions limit, NOT a Cloud Run Jobs limit (Cloud Run allows 24h).
+- **Secrets via Secret Manager** (never inline): `OPENAI_API_KEY`,
```
SimonKaran13, line 67: [P1] Specify the embedding configuration the jobs actually require

The checked-in backend example selects `EMBEDDING_PROVIDER=gemini` with `gemini-embedding-2`, but this planned secret/config list omits `GOOGLE_API_KEY` and all of `EMBEDDING_PROVIDER`, `EMBEDDING_MODEL`, `EMBEDDING_DIM`, and `COLLECTION_NAME`. Without them, a job can fall back to the OpenAI defaults and write a different vector space into the same environment-named collection. Define the complete immutable per-environment corpus configuration, keep only keys in Secret Manager, and validate its version/provenance before bootstrap, ingestion, retrieval, or snapshot restore.

SimonKaran13 (changes requested):
I reviewed the exact one-commit stack diff from `1b1ee121e81e9b69b4456a1029be80a7537009d4` to `7a19f735cd97b43d5dfab53656a3e3e7682a9a1b`. The immediate base is still #121's current head, so these findings are on #122 rather than restack noise.

I am requesting changes because two scaffolds that are explicitly usable today can mutate real data without preserving the guard they claim. Real Firestore seeding can combine the confirmed production project with the default dev fixture set, and Qdrant restore converts any failed preflight read into "collection missing" before a destructive `priority=snapshot` recovery. The removed München fixtures also do not remove the already-seeded München documents, so the PR's stated removal does not take effect on an existing environment.

I have not treated the deliberately absent Terraform, Workload Identity Federation, Cloud Scheduler, Cloud Run Job definitions, or one-time backfill as defects in this draft. Before this branch leaves draft, however, the rollout contract still needs an explicit development CLI runbook for the service and jobs: fixed project/region/environment/service names, API enablement, separate least-privilege service accounts, Secret Manager bindings, complete embedding/provider configuration, authenticated and anonymous health checks, corpus/source-period verification, and rollback. The current document only runs the image locally while pointing it at a remote Qdrant; it does not deploy the stack to development.

Validation on the exact head: `uv sync --locked` passed; the no-emulator seed guard passed; both Firestore seed integration tests passed against the local emulator; all six Firestore rules tests passed; the restore script passes `bash -n`; and `docker build --check ai-backend` reported no warnings. GitHub shows all five CI jobs passing, but each ran twice for the same PR SHA. The new seed tests are not actually covered by that workflow: one is explicitly ignored and the other skips when CI has no Firestore emulator. The workflow also does not exist on #114-#121, so their green checks remain Vercel-only. Current upstream releases are checkout v7.0.1, setup-uv v9.0.0 / uv 0.11.32, setup-node v7.0.0, setup-java v5.6.0, firebase-tools v15.24.0, and Qdrant v1.18.3; this workflow uses older action majors, unpinned firebase-tools via npm, and Qdrant 1.18.2.

#122 has no existing human review threads to duplicate. The stack remains transitively stale because #116 still records obsolete #115 base `8f0f1f9` while #115 now points at `199f9bd`; restack #116 through #122 after fixes.

PR #126, opened 2026-07-30
Feat/pdf manifesto ingestor

annikasiefke:
## Summary

Some Wahlprogramme are not covered by Abgeordnetenwatch (AW) (e.g. communal elections in general or a state election before AW opens its period). This adds a `manifesto_uploads` connector that allows us to ingest PDFs into the same `party_manifesto` corpus AW writes to (tagged `source="upload"`), plus the bucket-upload Makefile target that gets the PDFs there. `region`/`level`/`publish_date` are always derived from the Firestore fixtures for that election, never typed by hand.

**Not in this PR:** an actual Cloud Run Job/scheduler (see below).

---

## Running locally

```bash
make stores-up
make run-manifesto-uploads ARGS="--check"     # validate, no reads
make run-manifesto-uploads ARGS="--dry-run"   # parse + chunk, no embed cost
make run-manifesto-uploads                    # actually embed + upsert
make run-manifesto-uploads ARGS="--only spd,cdu"
```

Reads PDFs from `ai-backend/data/manifesto_uploads/{ENV}.txt` (preferring the staged copy at `firebase/storage_data/public/...`, else the bucket), and election metadata from the local seed files `firebase/firestore_data/{ENV}/*.json`.

## Uploading PDFs to the bucket

```bash
make upload-manifesto-uploads
make upload-manifesto-uploads ELECTION=landtagswahl-sachsen-anhalt-2026
make upload-manifesto-uploads ENV=prod   # prompts: Continue? [y/N]
```

One target, two steps that can't be half-done: `gcloud storage cp` to the bucket, then `gcloud storage objects update --add-acl-grant entity=AllUsers,role=READER`. Citations are plain `storage.googleapis.com` URLs governed by object ACLs (not `firebase/storage.rules`), so skipping the ACL step means every citation 403s. Defaults to the dev bucket.

## Running remote / how it maps to a future Cloud Run Job

Same target, different env vars — no separate script:

```bash
make run-manifesto-uploads QDRANT_URL=<deployed-host> QDRANT_API_KEY=... ENV=dev \
  MANIFESTO_UPLOADS_SOURCE=bucket ELECTION_FIXTURES_SOURCE=firestore
```

`MANIFESTO_UPLOADS_SOURCE=bucket` and `ELECTION_FIXTURES_SOURCE=firestore` switch both inputs from local files to the live bucket/database (same path the cloud run job will take in the future). 

---

## How the ingestor works

**The object path is the metadata.** A PDF at `public/{context_id}/{party_id}/{name}_{YYYY-MM-DD}.pdf` carries its election and party in the path, so they can't disagree with where the file lives. The filename date is the document's own date; the chunk's `publish_date` is the *election* date from the Firestore fixture instead, matching how AW manifestos are stamped.

**What's ingested from the bucket:** with `MANIFESTO_UPLOADS_SOURCE=bucket`, every object under `gs://<bucket>/public/` is listed and only those matching `public/{election}/{party}/{name}_{date}.pdf` are kept, i.e. other public assets (icons
etc.) are skipped, not fatal. If the bucket has content but *none* of it matches that shape, that's a hard failure rather than an empty list (a wrong bucket must never read as "retire everything"). Either backend (manifest or bucket) is a complete desired-state list, so **removing an entry retires its stored chunks on the next run**.

**De-dup with AW**:
1. Before ingesting an upload, skip it if AW already has a manifesto matching the same `party_id` + `region` + election `publish_date`.
2. If AW ingests that programme *later*, its `post_upsert` hook deletes the uploaded twin on the same three-field match after the AW copy is durably written.

A date mismatch just leaves a visible duplicate rather than deleting the wrong document.


ai-backend/src/ingestion/connectors/manifesto_uploads/bulk.py
```diff
@@ -0,0 +1,331 @@
...
+    # An invalid floor is a loud misconfiguration, raised before any work.
+    try:
+        since_floor = resolve_since_floor(args.since)
+    except ValueError as exc:
+        print(f"ERROR: {exc}", file=sys.stderr)
+        sys.exit(1)
+
+    manifest_path = Path(args.manifest).expanduser() if args.manifest else None
+    try:
+        entries = load_manifest(manifest_path, args.env)
```
SimonKaran13: [P1] The documented remote command does not use the selected work-list for its preflight. With `MANIFESTO_UPLOADS_SOURCE=bucket`, this still reads `dev.txt`; `ingest()` then re-discovers a different list from the bucket. New bucket-only objects therefore bypass the promised validation, while `ENV=prod` cannot start at all because this PR adds no `prod.txt` and exits at line 278 before listing the bucket. Resolve the preflight through `work_list(manifest_path, args.env)` (while still letting the connector discover stored retirements), and cover a bucket-backed production startup.

ai-backend/src/ingestion/connectors/manifesto_uploads/connector.py
```diff
@@ -0,0 +1,477 @@
...
+        """Read and parse one uploaded PDF, or flag it retired/skipped.
+
+        Failures return as ``skip_reason`` rather than raising (the runner doesn't
+        wrap fetch(); normalize() surfaces the skip instead).
+        """
+        object_path = external_id
+
+        # Not in the work-list → retire. Guarded on non-empty: an empty work-list
+        # means discovery found nothing, not "retire the whole uploaded corpus".
+        if self._expected_paths and object_path not in self._expected_paths:
```
SimonKaran13: [P1] This truthiness guard prevents the final document from ever being retired. I reproduced an empty manifest with one stored upload: `discover()` correctly returns the stored path as retired but leaves `_expected_paths=set()`; `fetch()` then tries the deleted PDF, returns `skip_reason`, and the runner preserves the old chunks. That contradicts both the complete-desired-state contract and the bucket-deletion promise. Once `discover()` has completed, any returned path absent from `_expected_paths` is a retirement even when the expected set is empty. Please cover an empty bucket plus an existing stored upload through the real runner.

ai-backend/src/ingestion/connectors/manifesto_uploads/connector.py
```diff
@@ -0,0 +1,477 @@
...
+                    must_not=[
+                        qdrant_models.FieldCondition(
+                            key="source",
+                            match=qdrant_models.MatchValue(value=self.source),
+                        )
+                    ],
+                ),
+                exact=True,
+            ).count
+        except Exception as exc:  # noqa: BLE001
```
SimonKaran13, line 459: [P2] Failing open here turns a transient count error into a permanent duplicate. I reproduced `count() -> error` producing upload chunks; on the next healthy run the AW copy is detected and `normalize()` raises, which deliberately preserves the just-written upload. Because the AW program is already complete, its `post_upsert` cleanup does not normally run again either. Treat an unavailable overlap check as a retryable per-item failure before writing rather than assuming “no AW copy,” and add this two-run regression.

ai-backend/src/ingestion/connectors/manifestos/supersede.py
```diff
@@ -0,0 +1,104 @@
...
+    # De-duplicate in case a future multi-party batch repeats the same triple.
+    targets = {
+        (c.party_id, c.region, c.publish_date)
+        for c in chunks
+        if c.source_type == SourceType.PARTY_MANIFESTO
+    }
+
+    superseded = 0
+    for party_id, region, publish_date in sorted(targets):
+        twin_filter = _upload_twin_filter(party_id, region, publish_date)
```
SimonKaran13: [P1] Party + region + election date does not identify the same document. The upload side deliberately supports multiple object paths for one party/election, but one AW program causes this filter to delete every one of them—for example, both halves of a split programme or a separate party document that AW does not carry. That is irreversible corpus loss, despite the stated preference for a visible duplicate over a wrong delete. Keep the safe duplicate until there is a document-level mapping, or require explicit operator retirement; add a regression with two uploads where only one has an AW replacement.

ai-backend/src/ingestion/connectors/manifestos/connector.py
```diff
@@ -555,3 +555,21 @@ def normalize(self, raw: dict) -> list[ChunkRecord]:
...
+        """Delete the uploaded twin of the programme just written, if any.
+
+        Runs only after the AW chunks are durably upserted, so the corpus is never
+        without the programme.
+        """
+        from src.ingestion.connectors.manifestos.supersede import (  # noqa: PLC0415
+            supersede_uploaded_manifestos,
+        )
+
+        return supersede_uploaded_manifestos(qdrant, collection_name, chunks)
```
SimonKaran13: [P1] This cleanup is a one-shot side effect after the AW chunks and their completion marker are already stored. If its count/delete call fails, the runner reports the AW ID as failed, but the next normal `discover()` sees that ID as complete and excludes it, so the uploaded duplicate is never reconciled without an operator knowing to force `MANIFESTO_REFRESH`. Make supersede independently retryable/idempotently discoverable (including already-complete AW parents), and inject a failure after the AW upsert to verify the next run finishes the cleanup.

Makefile
```diff
@@ -223,6 +224,49 @@ run-manifestos: bootstrap-collection
...
+	fi
+	@if [ -n "$(ELECTION)" ]; then \
+	  test -d "firebase/storage_data/public/$(ELECTION)" || \
+	    (echo "ERROR: firebase/storage_data/public/$(ELECTION) does not exist" && exit 1); \
+	  SRC="firebase/storage_data/public/$(ELECTION)"; SCOPE="public/$(ELECTION)/**"; \
+	else \
+	  SRC="firebase/storage_data/public/*"; SCOPE="public/**"; \
+	fi; \
+	echo "Uploading $$SRC -> gs://$(UPLOAD_BUCKET)/public/ ..."; \
+	gcloud storage cp -r $$SRC "gs://$(UPLOAD_BUCKET)/public/" && \
```
SimonKaran13: [P1] These two commands can be half-done: `cp` commits private objects before the separate ACL update, so an ACL permission/glob/auth failure leaves every citation at 403 while the files already exist. The current 12 manifest URLs all return 403 in the bucket dry run. `gcloud storage cp` supports `--predefined-acl=publicRead`; apply the ACL as part of each upload (or enforce an appropriate bucket default) and fail the target unless a representative final citation returns 200.

SimonKaran13 (changes requested):
I reviewed the exact `develop` `a897cfc` → `681fcba` diff (11 commits, 32 files) and am requesting changes for the six authored correctness issues below.

The overall direction is sound: uploaded PDFs reuse the shared chunker/runner, source IDs are isolated from AW, and the file/Firestore metadata gate prevents silently unreachable chunks. The unsafe parts are the lifecycle edges. The documented remote Make command validates the checked-in manifest instead of the selected bucket; an empty desired state cannot retire the final stored uploads; both AW overlap directions can leave permanent duplicates after transient Qdrant failures; and the destructive supersede identifies a “twin” only by party + region + election date, which can delete unrelated or split uploaded documents.

The upload command also does not provide its advertised “cannot be half-done” property: objects are copied before a separate ACL mutation. A read-only dry run against the 12 manifest URLs currently gets HTTP 403 for every file, which is exactly the intermediate state this sequence permits. This may simply mean the upload/ACL step has not been completed yet, but the current PR inputs cannot be parsed or cited from the bucket as documented.

The change adds roughly 2,300 lines of tests, but many of them restate implementation details while the empty-work-list and failure-injection cases below are absent. I would prune some of the duplicated prose/trivial assertions and spend that budget on a small number of end-to-end lifecycle regressions.

On the exact head, the full backend suite passed (805 passed, 2 skipped); the 292 focused uploaded/AW manifesto tests passed; Ruff passed over `src` and `tests`; mypy passed over `src`; and `git diff --check` is clean. GitHub currently exposes only successful Vercel checks, not the backend checks.

ai-backend/src/ingestion/connectors/manifesto_uploads/bulk.py
annikasiefke: preflight now goes through `work_list()`, so it validates the list actually ingested and bucket-backed `ENV=prod` starts. Also added a fix, so empty work-lists no longer hard-exit.  Covered by `test_bucket_backend_starts_for_an_env_that_ships_no_manifest`.

ai-backend/src/ingestion/connectors/manifesto_uploads/connector.py
annikasiefke: Guarded on `_discovered` (set only after `work_list()` returns) instead of on a non-empty set, so the last document is retirable. Covered through the real runner for both backends: `test_an_empty_bucket_retires_the_last_stored_upload`, `test_an_empty_work_list_retires_the_last_stored_upload`.

ai-backend/src/ingestion/connectors/manifesto_uploads/connector.py
annikasiefke: Raises `OverlapCheckUnavailable`, surfaced as a per-item failure, so nothing is written while the answer is unknown. Two-run regression added: `test_an_unavailable_overlap_check_writes_nothing_and_retries_next_run`.

ai-backend/src/ingestion/connectors/manifestos/supersede.py
annikasiefke: I changed this, but different than suggested. I think we should have some kind of automatic clean up, but agree that this was too generic. 
What I did now: Rather than keeping the duplicate, the class folder (introduced wahlprogramme subfolder per party, exclusively dedicated to Wahlprogramme which will overlap with Abgeordneten Watch sources) gives a document-level distinction without a mapping table: only `wahlprogramm` is eligible, and deletes address each document's deterministic `source_item_id` rather than the party+region+date filter used to find them (now discovery-only).

Makefile
annikasiefke: Single `cp --predefined-acl=publicRead`, separate ACL pass removed, and the target now fails unless a representative citation returns 200. Verified the flag and probe logic locally. I think the 403s were an empty bucket and not necessarily a result of this though, since I didn't run data ingestion of the new sources into the GCS yet, wanted to wait until PR is approved.

SimonKaran13 (approved):
lgtm

PR #127, opened 2026-07-30
feat: route Gemini chat + embeddings through Vertex AI, AI Studio as fallback

### LINKED ISSUE #124: integrate vertex ai service account

RobinFrasch:
## What this does

Routes Gemini **chat and embeddings** through Vertex AI so the spend lands on a dedicated GCP project, keeping Google AI Studio as an automatic fallback.

Auth is a service-account key rather than ADC/Workload Identity, and that is forced rather than chosen: the billing project enforces domain restricted sharing (`iam.allowedPolicyMemberDomains`), so binding our runtime identity there is rejected. A key minted *inside* the billing project is the only remaining path.

Closes #124.

## Key point: zero new dependencies

The already-installed `langchain-google-genai==4.2.4` speaks Vertex on both classes we use — passing a `credentials` object makes its `_determine_backend()` route to Vertex (`_common.py:566-582`), and `GoogleGenerativeAIEmbeddings` has the identical branch. No `langchain-google-vertexai`, no `google-cloud-aiplatform`, no `ChatVertexAI` migration, and `thinking_level` / `thinking_budget` carry over unchanged because it is the same class.

`pyproject.toml` and `uv.lock` are untouched.

## How it works

- **New** `src/google_credentials.py` resolves the key from `VERTEX_SA_JSON` (raw JSON, Cloud Run) or `VERTEX_SA_JSON_FILE` (path, local dev). Cached. **Absent** credentials return `None` without raising; **broken** ones (supplied but unusable) warn and return `None`, or raise under `VERTEX_REQUIRED` — see *Review round 2* at the end.
- **`src/llms.py`** gains a `_gemini()` helper (both backends, one constructor) and registers Vertex models at priority 192-200, above every AI Studio entry (max 100). The existing priority-sorted failover in `get_answer_from_llms` / `get_structured_output_from_llms` / `stream_answer_from_llms` therefore tries Vertex first and falls through to the *identical* AI Studio model on any error — **no change to those functions was needed**.
- **`src/embeddings.py`** gains a Vertex branch inside the existing `gemini` block, with `EMBEDDINGS_USE_VERTEX=0` as a manual kill-switch.

### Three deliberate design decisions worth reviewing

1. **`EMBEDDING_PROVIDER` stays `"gemini"`.** It names the *vector space*, not the transport, and it is stamped into the Qdrant embedding-space fingerprint that `setup_collection.check_fingerprint()` enforces. Encoding transport there would reject the existing corpus and force a full re-ingest. Verified safe: Vertex and AI Studio return **bit-identical vectors** for the same input (cosine `1.00000000`, max abs diff `0.0`). `setup_collection.py` is untouched.

2. **`GOOGLE_APPLICATION_CREDENTIALS` is never set.** `src/firebase_service.py` falls back to bare `initialize_app()` (ADC), so exporting the Vertex key globally would make **firebase-admin authenticate as the wrong project**. Credentials are passed only as an explicit `credentials=` argument.

3. **`vertexai=False` is pinned on the AI Studio path.** Backend detection consults `GOOGLE_GENAI_USE_VERTEXAI` *before* defaulting to AI Studio, so if that variable ever leaked in, every fallback client would silently flip to Vertex and fail for want of credentials — collapsing the fallback exactly when needed.

## Findings from live testing

The first commit was untested; live calls corrected several assumptions.

**Model availability on the billing project is narrow:**

| model | europe-west3 | global |
|---|---|---|
| `gemini-2.5-flash` | OK | OK |
| `gemini-2.5-flash-lite` | 404 | OK |
| `gemini-3-flash-preview` | 404 | OK |
| `gemini-embedding-2` | 404 | OK |
| `gemini-2.0-flash` | 404 | 404 |

So `VERTEX_LOCATION=global` is **load-bearing**, not a default-by-convenience: a 404 falls through to AI Studio silently, so a regional value would quietly stop most traffic being billed to Vertex with no visible error. `gemini-2.0-flash` is not registered on the Vertex tier at all. `global` also matches the prior AI Studio behaviour (no regional pinning either), so this is not a residency regression.

## Notes for the reviewer

Two pre-existing issues surfaced during testing. **Neither is caused by this PR and neither is fixed here** — flagging rather than scope-creeping:

- **`gemini-embedding-2` ignores `task_type` entirely.** `RETRIEVAL_QUERY` and `RETRIEVAL_DOCUMENT` return bit-identical vectors. Confirmed at the raw `google-genai` SDK level *and* on AI Studio, so it is the model's behaviour. The document/query asymmetry that `embeddings.py` documents and that the ingestion/retrieval paths carefully pass has no effect. `gemini-embedding-001` did honour it, so this looks like a change in `-2`. Worth a separate issue.
- **`tests/test_ids.py:120-138`** hard-asserts the OpenAI embedding defaults (`text-embedding-3-large` @ 3072) while the live `.env` is `gemini-embedding-2`. It passes only in a clean env where `.env` has not leaked into pytest. Stale relative to the live config.

Also: `firebase/.emulator-data/` was committed by accident in 83e5c5b and is untracked here, with an ignore rule added. Local emulator scratch state — anonymous test accounts and a Firestore dump, no real credentials.

### Anyone can do this — no GCP access needed

The interesting behaviour is the **absence** path, and it is fully testable locally because the credential resolver degrades instead of raising.

```bash
cd ai-backend
uv run pytest -q          # 655 passed, 2 skipped
uv run ruff check src tests
uv run mypy src/google_credentials.py src/llms.py src/embeddings.py
```

The suite is hermetic — no `load_dotenv` in `tests/`, no `env_files` in the pytest config — so it exercises the `VERTEX_AVAILABLE=False` path with no credentials present. Specifically worth reading:

- `tests/test_google_credentials.py` — absent, malformed, and blank-value env all resolve to `None` rather than raising. **This is the property that matters**: a raise here would take down `src/llms.py` import and with it the whole service.
- `tests/test_llms_vertex.py` — without credentials the LLM lists are unchanged, the AI Studio clients are pinned `_use_vertexai=False`, and every Vertex priority exceeds every AI Studio priority.
- `tests/test_embeddings.py` — the new Vertex cases use a fake credentials object, so they assert the routing without a key.

You can also confirm the no-credentials path by inspection:

```bash
cd ai-backend
uv run python -c "from src import llms; print(llms.VERTEX_AVAILABLE, len(llms.RESPONSE_GENERATION_LLMS))"
# -> False 7   (identical to before this PR)
```

**Note:** import `src.llms`, not `src.google_credentials` directly — the latter deliberately does not call `load_env()`, so a standalone import reads only the process env and reports `False` even when `.env` is configured.

### Requires access to the billing GCP project (@RobinFrasch only)

Reviewers cannot verify the live half. It has been run and is recorded here for completeness:

```
VERTEX_AVAILABLE: True
200 vertex-gemini-3.0-flash-preview    OK  tokens=28  loc=global
195 vertex-gemini-2.5-flash            OK  tokens=3   loc=global
200 vertex-gemini-2.5-flash-lite-det   OK  tokens=3   loc=global
195 vertex-gemini-2.5-flash-det        OK  tokens=25  loc=global
RETRIEVAL_QUERY / RETRIEVAL_DOCUMENT   vertex=True  dim=3072
vertex vs ai-studio vectors: identical (cosine 1.00000000, max diff 0.0)
```

To re-run it, with the SA key placed at `ai-backend/vertex-sa.json` (gitignored) and `VERTEX_SA_JSON_FILE` / `VERTEX_PROJECT_ID` set in `ai-backend/.env`, then confirm billing actually moved:

```bash
gcloud logging read 'protoPayload.serviceName="aiplatform.googleapis.com"' \
  --project=<billing-project> --freshness=10m --limit=20
```

Traffic there means it worked. Empty, while chat still responds, means it fell through to AI Studio — check app logs for `Error invoking LLM vertex-…`.

### Deployment state

Secret Manager and Cloud Run wiring for `wahl-chat-backend-dev` is **already done** (secret stored in the app project, runtime SA granted `secretAccessor`, `VERTEX_SA_JSON` injected via `--set-secrets`). Still outstanding:

- **Prod (`wahl-chat`)** needs its own secret and binding — the secret is app-project-local by design, to avoid the cross-project IAM binding that the org policy would reject.
- **The ingestion Job does not exist yet**, so document embeddings still bill AI Studio. Apply the same two flags when it is created, or that path stays on the old billing route.
- `:latest` resolves at instance start, not per request — rotating the key needs a new revision to take effect.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Review round 2 (commit 9cb4679)

All four of @PaulLampe's comments were valid; each is answered inline on its thread. What changed:

**`vertexai` is now pinned on *both* paths.** Backend resolution puts `GOOGLE_GENAI_USE_VERTEXAI` *above* the "credentials imply Vertex" inference, so the Vertex path had the mirror image of the bug the AI Studio pin already guarded: `=false` would have dropped every Vertex client onto AI Studio carrying no API key. Applies to chat and embeddings — same field, same `_determine_backend`.

**Silent credential failures are gone.** Three paths returned `None` with nothing logged: blank `VERTEX_SA_JSON` (fell through both branches), a nonexistent `VERTEX_SA_JSON_FILE`, and credentials that load but yield no project id. The third was not in the review. The two cases are now split:

- *absent* (no `VERTEX_*` source) — quiet, `DEBUG` only, never raises. CI and local dev.
- *broken* (supplied but unusable) — always `WARNING`, naming the variable and, for the file form, the resolved **absolute** path.

**New `VERTEX_REQUIRED`**, off by default, escalates the broken case to `VertexConfigError` at import. It escalates *broken* config only — set alone, with nothing configured, it still cannot take the service down.

> ⚠️ **Do not set `VERTEX_REQUIRED=1` on prod until prod's secret is wired.** An unset variable is safe, but a secret that exists and is empty counts as broken and will fail the revision's import. Order: wire the secret → confirm the `Vertex AI enabled` log line → then enable.

**`vertex_enabled()` is no longer test-only** — it is the single decision point for "is Vertex on" in both `llms.py` and `embeddings.py`, replacing the duplicated two-part condition. In `embeddings.py` the `EMBEDDINGS_USE_VERTEX` check deliberately runs *first*, so opting out never touches the resolver.

**`llms.py` logs which backend it came up on at import.** Without it, the only symptom of a Vertex tier that failed to register is billing that never moves.

Verification: **668 passed** (was 655), ruff and mypy clean. Every new warn/raise path exercised from a shell. All four Vertex chat models and the embedding client re-invoked live after the pin (`_use_vertexai=True`, dim 3072). The `gcloud logging` billing confirmation was **not** re-run this round — auth had expired — but the live calls resolved against the billing project.


ai-backend/src/llms.py
```diff
@@ -8,13 +8,62 @@
...
+    if vertex:
+        return ChatGoogleGenerativeAI(
+            model=model,
+            max_retries=0,
+            credentials=_VERTEX_CREDS,
+            project=_VERTEX_PROJECT,
+            location=vertex_location(),
+            # Surfaces in Cloud Billing reports for per-app attribution.
+            labels={"app": "wahl-chat", "tier": "vertex"},
+            **kwargs,
```
PaulLampe, line 83: shouldn't we pin  vertexai=True here? or would that be unnecessary?

ai-backend/src/embeddings.py
```diff
@@ -97,16 +121,50 @@ def get_embeddings(
...
+        # derive `project` from the credentials object — it must be passed.
+        project = vertex_project() if credentials else None
+        if credentials is not None and project is not None:
+            return GoogleGenerativeAIEmbeddings(
+                model=resolved_model,
+                output_dimensionality=resolved_dim,
+                task_type=task_type,
+                credentials=credentials,
+                project=project,
+                location=vertex_location(),
```
PaulLampe, line 163: Same here?  vertexai=True

ai-backend/src/google_credentials.py
```diff
@@ -0,0 +1,120 @@
...
+# a data-residency regression. It is also the only location on the billing
+# project where gemini-embedding-2 and the 2.5-lite / 3-preview chat models
+# resolve; regional endpoints (europe-west3/4, us-central1) 404 on all of them
+# and serve gemini-2.5-flash alone. Revisit if EU residency becomes a hard
+# requirement — that would mean europe-west3 plus a much narrower model set.
+_DEFAULT_LOCATION = "global"
+
+
+@lru_cache(maxsize=1)
+def get_vertex_credentials() -> Optional[Any]:
```
PaulLampe, line 101: Is this the behavior we want? Should we fall back to AI Studio when we can't get the credentials? Or should this fail a bit louder than just log a warning?

Additionally, if the path set for VERTEX_SA_JSON_FILE is just wrong or VERTEX_SA_JSON is empty, this would fail completely silently, even without a warning log, as os.path.exists(path) returns false.

ai-backend/src/google_credentials.py
```diff
@@ -0,0 +1,120 @@
...
+
+    An empty or whitespace-only VERTEX_LOCATION falls back to the default rather
+    than being passed through: a blank region would otherwise reach the client
+    and fail obscurely, and `VERTEX_LOCATION=` is an easy thing to leave behind
+    in a shell or a Cloud Run env spec.
+    """
+    return os.getenv("VERTEX_LOCATION", "").strip() or _DEFAULT_LOCATION
+
+
+def vertex_enabled() -> bool:
```
PaulLampe, line 201: this is only ever referenced from tests, right?

PaulLampe (commented):
Minor comments, mostly edge cases

ai-backend/src/llms.py
RobinFrasch: Pinned `vertexai=True` and rewrote that docstring paragraph to state both directions. New `test_vertex_clients_are_pinned_on_vertex` covers it, and `test_ai_studio_clients_are_pinned_off_vertex` now asserts `vertexai is False` and not just the resolved `_use_vertexai` — the resolved value reads False with the kwarg dropped too, so on its own it never proved the pin was there.

ai-backend/src/embeddings.py
RobinFrasch: Yes — `GoogleGenerativeAIEmbeddings` has the same `vertexai: bool | None` field (`embeddings.py:154`) and a byte-identical `_determine_backend`, so the same gap applied. Pinned `vertexai=True`, asserted in `test_gemini_routes_to_vertex_when_credentials_present`.

ai-backend/src/google_credentials.py
RobinFrasch: Fixed.

Rather than delete it I made it the real one: it is now the single decision point for "is Vertex on", used by `llms.py` (`VERTEX_AVAILABLE = vertex_enabled()`) and by `embeddings.py`, replacing the two-part `credentials is not None and project is not None` condition that was duplicated in both. Free to call — the credential lookup is cached and the project lookup is an env read.

ai-backend/src/google_credentials.py
RobinFrasch: Both cases named were real, and there was a third: credentials that load but resolve no project id also disabled Vertex silently. Now split into absent config (no VERTEX_* at all — stays quiet, that's CI and local dev) and broken config (always warns, naming the variable and, for the file form, the resolved absolute path — wrong-CWD being the likely mistake).

On failing louder: added VERTEX_REQUIRED, off by default, which escalates the broken case to a raise at import. Meant for deployed revisions, where quiet fallback means invisible mis-billing. Kept off by default because an import-time raise takes the whole service down, which can't be the default for an optional key. It escalates broken config only — set alone with nothing configured, it still can't take the service down.

One case I decided the other way: a blank VERTEX_SA_JSON beside a valid VERTEX_SA_JSON_FILE warns but does not disable Vertex, since there Vertex is usable

PR #128, opened 2026-08-06
User-directed source-type filtering

annikasiefke:
## Summary

Source filtering feature: users can ask for specific source kinds directly (e.g. "Zeigt mir Videoaufnahmen zum Thema Lohnniveau", "Abstimmungen zur Infrastruktur").

A new LLM preprocessing step (concurrent with party targeting) detects the requested source kinds (`manifesto` / `votes` / `speeches` / `videos`, **default: all**; fail-open). Retrieval then only queries those legs, with raised top-k budgets at the unchanged score threshold; `videos` filters the speech leg to `source=="op"` in Qdrant. Filtered turns bypass the cross-user cache. Works in single-party and comparison chats; follow-ups keep the filter.

Also in this PR:

- **Video pills**: citations that open a recording render as a clapperboard "Video N" button instead of a bare number (chat + share pages).
- **Bund/Land separation** (prompt-only): Landtagswahl answers must signpost level switches, never "lehnt X ab [Bund]. Stattdessen Y [Land]" as one storyline.

## Trying it

`make dev`, Bundestagswahl chat: "Zeigt mir Videoaufnahmen zum Thema Lohnniveau" → video-only answer with pills; "Wie steht ihr zu Videoüberwachung?" → normal answer (negative control). Backend log shows each verdict: `User source filter active … ['videos']`.

### DIFF
--- a/ai-backend/src/chat_service.py
+++ b/ai-backend/src/chat_service.py
@@ -147,12 +147,8 @@ _CURRENT_MANIFESTO_LIMIT = 4
 _CURRENT_SPEECH_LIMIT = 2  # normal cap (ranked last)
 _CURRENT_SPEECH_FALLBACK = 5  # adaptive ceiling when official data is sparse
 
-# User-requested source-type filter (detect_source_filter): when the user
-# explicitly scopes the answer to specific source types, the requested sources
-# carry the WHOLE answer, so their current-bucket budgets are raised. The score
-# threshold is unchanged — the extra slots only fill with results that clear the
-# same relevance bar. The adaptive speech trim is skipped under a filter (an
-# explicitly requested source is never demoted to "supporting material").
+# Raised current-bucket budgets under a user source filter — the requested
+# sources carry the whole answer. Score threshold unchanged.
 _FILTERED_VOTE_LIMIT = 10
 _FILTERED_MANIFESTO_LIMIT = 8
 _FILTERED_SPEECH_LIMIT = 6
@@ -560,12 +556,10 @@ async def _retrieve_party_buckets(
     concurrently and a single source failure degrades to empty. WAHL_CHAT_PARTY has no
     corpus data → all buckets empty. log_prefix labels the failure logs per caller.
 
-    source_filter (detect_source_filter values) restricts retrieval to the
-    user-requested source types: only requested legs run (the others return empty
-    buckets untouched) and each requested leg gets the raised _FILTERED_* budget.
-    "videos" runs the speech leg with a Qdrant-level source=="op" provenance
-    filter so only video-bearing speeches return — a post-fetch filter would
-    starve videos whenever transcript-only speeches dominate the top-k.
+    source_filter restricts retrieval to the user-requested source types: only
+    requested legs run, each with the raised _FILTERED_* budget. "videos" adds a
+    Qdrant-level source=="op" filter — post-fetch filtering would starve videos
+    whenever transcript-only speeches dominate the top-k.
     """
     if party.party_id == WAHL_CHAT_PARTY.party_id:
         return _RetrievedBuckets([], [], [], [], [], [])
@@ -899,8 +893,8 @@ async def fetch_party_response_stream(
             # bar" — enough to instruct a marked historic section. Drives has_historic
             # threaded into generation below.
             has_historic = bool(manifesto_historic or speech_historic or vote_historic)
-            # Adaptive speech trim — skipped under a user source filter: an
-            # explicitly requested source keeps its raised (filtered) budget.
+            # Trim skipped under a source filter — requested speeches keep
+            # their raised budget.
             if not source_filter and not votes_absent and not manifesto_absent:
                 speech_current = speech_current[:_CURRENT_SPEECH_LIMIT]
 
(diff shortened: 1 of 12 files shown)


ai-backend/src/chat_service.py
```diff
@@ -1588,11 +1677,16 @@ async def generate_chat_stream(  # type: ignore[no-untyped-def]
                 # quick_replies (forgeable). First turn uses the server-verified
                 # proposed-question check; follow-ups are gated against the
                 # quick_replies the server recorded last turn for this session.
-                group_chat_session.is_cacheable = _evaluate_cache_eligibility(
-                    body.session_id,
-                    user_message.content,
-                    is_beginning_of_chat,
-                    is_proposed_question,
+                # Filtered turns are never cacheable — the cache entry doesn't
```
RobinFrasch, line 1680: Why are filtered turns not cacheable? Shouldn't it be possible to cache as usual the user questions with the generated return answer? Assuming the LLM-detected filter is not completely broken, where would be the downside?

ai-backend/src/prompts.py
```diff
@@ -183,6 +184,47 @@ def get_party_vote_behavior_summary_guidelines():
 SOURCE_LABEL_VOTES_DE = "namentlichen Abstimmungen"
 SOURCE_LABEL_SPEECHES_DE = "Reden"
 
+# =============================================================================
+# User-requested source-type filter (organic "zeig mir nur Videos/Abstimmungen/…")
+# =============================================================================
+# Nominative labels, keyed by the classifier's SourceFilterLiteral values.
+SOURCE_FILTER_LABELS_DE = {
```
RobinFrasch, line 191: Would it make sense to distinguish between Wahlprogramme of different years? Could be a benefit if people want to compare themselves how anything changed in the promises over the years.

ai-backend/src/prompts.py
```diff
@@ -477,6 +520,55 @@ def get_party_vote_behavior_summary_guidelines():
...
+- "speeches": Reden im Bundestag — was Abgeordnete der Partei im Plenum gesagt haben (z.B. „Reden", „Redebeiträge", „was wurde im Parlament gesagt").
+- "videos": Reden, zu denen eine Videoaufnahme existiert — der Nutzer will etwas ansehen/anhören (z.B. „Videos", „Videoaufnahmen", „Aufzeichnungen", „etwas zum Anschauen").
+
+# Aufgabe
+Gib die Liste der Quellenarten zurück, auf die der Nutzer die Antwort ausdrücklich beschränken will. Mehrere Quellenarten sind möglich (z.B. "Abstimmungen und Reden zum Thema Infrastruktur" → ["votes", "speeches"]).
+
+## Wichtige Hinweise zur Einstufung
+- Standard ist die LEERE Liste: Die allermeisten Nachrichten fragen nach einem THEMA, nicht nach einer Quellenart. Gib nur dann Quellenarten an, wenn der Nutzer sie erkennbar verlangt.
+- Nutzer verwenden beliebige eigene Formulierungen — auch Umschreibungen, Umgangssprache, Tippfehler oder englische Begriffe. Entscheide anhand der BEDEUTUNG der Nachricht, welche Quellenart gemeint ist; die Formulierungen in den Hintergrundinformationen sind nur Beispiele, keine vollständige Liste.
```
RobinFrasch, line 539: You have defined in another file that if the user requests speeches, the filter allows for plenarprotokolle AND videos. Isn't that then a contradiction here?

RobinFrasch (commented):
Tested locally, nice behaviour! Some comments/questions about caching and filtering logic, not blocking. Can be moved to a separate PR if out of scope, except the potential conflicting information (ref. comment in line 539)

ai-backend/src/chat_service.py
annikasiefke: The cache key (message text / history hash) doesn't encode the detected filter. Since the classifier is an LLM call identical text can classify differently across runs, so I guess in most cases it should work, but given it's non-deterministic, answers might not get the same filters across all users -> with cross-user first-write-wins replay, a ['videos'] answer could be served to a turn classified as [], or vice versa. 

I can either enable caching and we take that "risk" or we can fold the filter into the cache key or we leave as is for now, your call! Only cacheable answers are currently answers to the proposed questions or quick replies - proposed questions should generally not be source scoped, quick replies might be depending on what follow-up questions the LLM cam e up with.

ai-backend/src/prompts.py
annikasiefke: It partially exists with the historical context section of the answer. If we do what you suggest, I would remove the historical context from the "default" answer and make it a second filter axis, where we filter by wahlperiode, so we can have explicit answers to questions like "wie hat sich eure Position seit 2021 verändert?". Should I do that here or as a follow up?

RobinFrasch: then lets put that on the todo list as a follow up and discuss priorities

ai-backend/src/chat_service.py
RobinFrasch: then lets keep the cache at proposed questions and follow up only for now and consider that an avenue to save funds later if necessary

RobinFrasch (approved):
All comments fixed or moved to future work

PR #135, opened 2026-08-08
fix(cache): scope party answer cache by election context

annikasiefke:
## Problem
Cached party answers were keyed only by party + question (`cached_answers/{party_id}/{cache_key}`). Party ids repeat across elections, so clicking the proposed question "Was kann ich fragen?" in the Sachsen-Anhalt chat replayed the cached Bundestagswahl-2025 answer. The multi-turn history-hash keys collided the same way.

## Fix
- Cache docs now live under `cached_answers/{context_id}__{party_id}/{cache_key}` — both Firestore helpers take a required `context_id`, passed from `GroupChatSession.context_id`.
- Removed a dead 5-segment `firestore.rules` match (odd segment count never matches a document); the new composite doc id is covered by the existing rule.

## Notes
- Legacy un-scoped cache docs are deliberately **not** migrated: the collision means they mix contexts and can't be attributed. They're orphaned; each election's cache refills lazily. Optional follow-up: one-off deletion of the old docs.
- Tests: full backend suite + SSE smoke green; the cache-write test now asserts the context id.

### DIFF
--- a/ai-backend/src/chat_service.py
+++ b/ai-backend/src/chat_service.py
@@ -815,7 +815,9 @@ async def fetch_party_response_stream(
             )
             existing_cached_answers: List[
                 CachedResponse
-            ] = await aget_cached_answers_for_party(party.party_id, cache_key)
+            ] = await aget_cached_answers_for_party(
+                group_chat_session.context_id, party.party_id, cache_key
+            )
             cached_answer_limit = 1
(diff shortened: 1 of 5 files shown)


romman8:
superseded by https://github.com/wahl-chat/wahl-chat-app/pull/154

PR #138, opened 2026-08-08
feat(seo): serve a real page at / instead of redirecting by IP

RobinFrasch:
`/` was a 307 to whichever election geo-IP picked, so it never returned content. Every inbound backlink — tum.de, cit.tum.de, kas.de, politikzumanfassen.de, go-vote.de — landed on a redirect, and the destination moved every time an election concluded, so authority never accumulated anywhere. Googlebot crawls from US IPs, so it was also sent somewhere different from a German visitor.

`/` now renders the same page for everyone. The featured election is derived from election dates via `lib/elections.ts` rather than from the request, so there is nothing left to vary by IP and nothing to go stale — `REGION_TO_CONTEXT` and `getContextIdFromGeo` are gone.

Layout matches a context page with two differences: the party-logo grid becomes a slowly drifting colour field that links into the featured election, and the heading says what the site is for. Below the fold, intro prose and links to every election, so the URL holding the site's inbound links passes them on.

Verified crawler and browser receive byte-identical `<head>` and visible text.


RobinFrasch:
Superseded by #163 — same work, rebased onto develop. This PR could not be retargeted after #137 merged: GitHub refuses base changes on stacked PRs, and the stack could not be dissolved because a merged PR cannot leave it, so the diff here would have carried all 61 develop commits.

PR #139, opened 2026-08-08
fix: unbreak the BUDGET_SPENT kill switch, correct Context.date type

RobinFrasch:
- `BUDGET_SPENT` was checked before the static-page check, and `/budget-spent` is itself a static page — the kill switch redirected its own landing page to itself forever, and sent `/sitemap.xml`, `/robots.txt` and `/manifest.json` there too
- `Context.date` was declared `string | null` but `firebase-server.ts` maps it through `firestoreTimestampToDate()`; `formatGermanDate` only worked on it by accident
- `/chat` and `/session` now 308 to `/`, which is a stable target; paths carrying a party or session id still need a context and stay 307
- deletes `app/(with-header)/swiper/page.tsx` — unreachable since middleware intercepts `/swiper`, still shipping a 407 kB bundle
- drops the unconditional 500 ms `setTimeout` that forced skeleton party cards even though the server layout had already supplied the parties
- adds unit tests for the election date rule (5-day buffer, undated and unparseable dates, all-concluded list) using the Bun runner, so no test framework is pulled in


RobinFrasch:
Superseded by #164 — same work, rebased onto develop as part of rebuilding this stack. See #138 for why the old stack had to be recreated.

PR #144, opened 2026-08-08
add: update how to for current features and dynamic display of served…

### LINKED ISSUE #143: update how to page with current information
- make elections shown dynamic
- add information about more data sources

RobinFrasch:
# Update how-to page with current information (#143)

## Summary
The "Welche Wahlen werden unterstützt?" section listed elections in a hardcoded array, so it went stale after every election day (it still advertised BW/RP/München 2026 as upcoming and referenced a "Bundesebene" area that no longer exists). This PR loads that list from the live contexts instead, and refreshes the surrounding copy to reflect the sources we actually use today.

### Dynamic election list

HowTo takes an optional contexts prop and otherwise falls back to useElectionContext({ optional: true }) — the standalone /how-to page passes them in from the server, while HowToDialog inside a chat reads them from the existing ContextProvider.
buildElectionsContent() builds the panel at render time: filters to type === 'election', splits into Kommende Wahlen / Vergangene Wahlen, formats dates with formatGermanDate, and drops the intro/outro sentences that no longer apply. A type === 'general' context (none exist today) would get its own group.
/how-to is now an async page fetching getContexts() with revalidate = 3600.
Extracted splitContextsByElectionDate() into lib/utils.ts so the 5-day post-election buffer lives in one place; election-select.tsx now uses it instead of its own inline copy.
The dynamic list also flows into the PDF export (no exporter changes needed — it reuses the sections shape).

### Content updates

Process steps and the data-sources section now name all four source types — Wahlprogramme, Abstimmungen (abgeordnetenwatch.de), Plenarprotokolle (DIP des Bundestags) and Debattenvideos (OpenParliamentTV) — each linked via a new SOURCE_LINKS map plus a LinkedText helper that replaced the hardcoded abgeordnetenwatch-only special case.
"Abstimmungsverhalten analysieren" explains that the abgeordnetenwatch data feeds answers by default, not just via the button.
"Nach welchen Kriterien werden die Parteien ausgewählt?" now states we aim to cover every party standing for election that has public information on its plans, and points at [info@wahl.chat](mailto:info@wahl.chat) for missing ones.

### DIFF
--- a/web/components/how-to.tsx
+++ b/web/components/how-to.tsx
@@ -1,6 +1,12 @@
 'use client';
 
-import { exportHowToPDF } from '@/lib/how-to-pdf-export';
+import type { Context } from '@/lib/firebase/firebase.types';
+import {
+  type HowToAccordionContent,
+  type HowToAccordionItem,
+  exportHowToPDF,
+} from '@/lib/how-to-pdf-export';
+import { formatGermanDate, splitContextsByElectionDate } from '@/lib/utils';
 import {
   DownloadIcon,
   MessageCircleQuestionIcon,
@@ -11,10 +17,11 @@ import {
   WaypointsIcon,
 } from 'lucide-react';
 import Link from 'next/link';
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 import ChatActionButtonHighlight from './chat/chat-action-button-highlight';
 import { MAX_SELECTABLE_PARTIES } from './chat/chat-group-party-select-content';
 import ProConIcon from './chat/pro-con-icon';
+import { useElectionContext } from './providers/context-provider';
 import {
   Accordion,
   AccordionContent,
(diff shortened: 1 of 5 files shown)


web/components/how-to.tsx
```diff
@@ -53,13 +60,76 @@ const INTRO_TEXT = {
...
+      'Über die Wahlauswahl auf der Startseite kannst du jederzeit zwischen den Wahlen wechseln.',
+  ]
+    .filter(Boolean)
+    .join(' ');
+
+  return {
+    intro:
+      sections.length > 0
+        ? 'Welche Wahlen wir unterstützen, richtet sich nach dem Wahlkalender. Aktuell sind das:'
+        : 'Die kommenden Wahlen haben wir noch nicht integriert. Bis diese verfügbar sind, stöber doch bei den vergangenen Wahlen nach interessanten Inhalten.',
```
annikasiefke, line 126: I think `sections.length === 0` only holds when contexts is completely empty (happens, e.g., on Firestore error: `getContexts()` returns `[]` and is cached for 1h via `revalidate = 3600`), so this tells users to browse past elections while rendering none. 

Suggestion: move that sentence into `outro` guarded on `upcoming.length === 0 && past.length > 0`, and make the empty-contexts intro something neutral like "Die aktuell unterstützten Wahlen findest du in der Wahlauswahl auf der Startseite."

annikasiefke (approved):
One minor, non-blocking thing about the wording on the how-to page. But it's a nit, branch already lgtm!

PR #146, opened 2026-08-09
feat(pdf): in-page PDF viewer with cited-page jump and quote highlighting

annikasiefke:
Citations now open in an in-page pdf.js viewer instead of the browser's native one, so jumping to the cited page works on every device, including mobile. The backend sends a `snippet` of the cited chunk with manifesto and speech sources, and the viewer highlights that passage (fuzzy match, tolerant of hyphenation) and opens centered on it:

image

The document scrolls continuously like a native viewer, with windowed rendering so large manifestos stay cheap on phones; GCS-hosted PDFs are fetched directly, Bundestag/AW PDFs keep using the proxy. Pinned to **react-pdf 9** / **pdfjs-dist 4** because **pdfjs-dist 5.4.x** crashes under Next's bundled **webpack** (webpack#20095). Also includes LAN-testing wiring (emulator/backend binds, secure-context UUID fallback), so testing locally on phones via network address works as well.

Verified on Safari (macOS + iOS), Chrome (macOS).

### DIFF
--- a/web/components/chat/pdf-viewer.tsx
+++ b/web/components/chat/pdf-viewer.tsx
@@ -191,6 +191,21 @@ function PdfViewer({
     setCurrentPage(page);
   }, [slotHeight, numPages]);
 
+  // Missing highlights are silent for users by design, but the two causes need
+  // to stay distinguishable for developers (a source without a snippet vs. a
+  // snippet the page no longer matches, e.g. after a corpus re-chunk). Logged
+  // once per viewer instance — the text layer re-renders on zoom/resize.
+  const highlightStateLogged = useRef(false);
+  const debugHighlightState = useCallback(
+    (message: string, detail?: Record<string, unknown>) => {
+      if (!highlightStateLogged.current) {
+        highlightStateLogged.current = true;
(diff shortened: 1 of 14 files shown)


web/components/chat/pdf-viewer.tsx
```diff
@@ -0,0 +1,393 @@
...
+  }, [slotHeight, numPages]);
+
+  // TextItem vs TextMarkedContent: only real text items carry `str`; marked-
+  // content entries still occupy indices, so keep positions aligned with ''.
+  // transform[5] is the item's y in PDF text space — the matcher uses it to
+  // keep visually distant matches (running footers) out of the passage cluster.
+  const onTextSuccess = useCallback(
+    (textContent: { items: unknown[] }) => {
+      if (!snippet) {
+        return;
```
RobinFrasch, line 216: I encountered an issue here while testing, because I had an outdated backend: A missing snippet field and a snippet that doesn't match the page produce identical behaviour. A debug that distinguishes "no snippet on this source" from "snippet present, 0 matches" could be helpful, especially if we work with a continuously changing dataset in the future.

RobinFrasch (approved):
One small comment, then ready

PR #154, opened 2026-08-16
Key proposed-question answers on the full LLM request

### LINKED ISSUE #92: fix: make caching context dependent
Image

Current caching only checks for exact text and not for the context in which this message is sent

romman8:
## Why

We cache answers to proposed questions so we do not call the answer LLM again for the same request. The cache key must include every input to that LLM except the clock. If we key only on the question, we can serve an old answer after a prompt, model, or corpus change. The RAG query rewrite is also an LLM call. If we do not cache it, the same question can get different chunks and miss the answer cache.

## What this does

- Build the answer cache key from the full LLM request. Do not include date or time.
- Encode each field with its byte length so two different values cannot share a key.
- Put retrieved chunks in the key. A new source causes a miss. The same sources in a different order still hit. A change to URL or page causes a miss.
- For wahl.chat, also put context name, date text, location, and the party list in the key.
- Cache the RAG rewrite on curated turns. A repeat uses the stored query.
- Look up the answer after retrieval. A hit skips the answer LLM. Only curated chats use the cache.

## Test plan

- [ ] `cd ai-backend && uv run pytest tests/test_answer_cache.py tests/test_chat_service.py tests/test_chat_sse.py tests/test_chatbot_async.py tests/test_source_filter.py`
- [ ] First-turn proposed question writes 64-char hex keys for the rewrite and the answer
- [ ] Same question after a prompt, model, URL/page, or party-list change misses
- [ ] Same sources in a different order still hit; a new source misses
- [ ] A free-text turn writes neither cache


ai-backend/src/answer_cache.py
```diff
@@ -0,0 +1,138 @@
...
+
+def canonicalize_rag_context(docs: list[Document]) -> str:
+    """Serialize retrieved chunks in a retrieval-order-independent form.
+
+    Same set of sources → same string, even if HNSW returned them in a
+    different rank order. A new or dropped source changes the set and
+    therefore the cache key, so the next user gets a freshly generated
+    answer grounded on the updated corpus.
+    """
+    return get_rag_context(sorted(docs, key=_rag_doc_sort_key))
```
SimonKaran13: The URL and page are only used for sorting here; they are not included in the text that gets hashed. If a link or page changes, the cache may still return the old source. Could we include them in the cache key?

SimonKaran13 (approved):
Looks good, let's merge after resolving the comments

PR #156, opened 2026-08-23
fix(firebase): revalidate the live site cache after a real-project seed

romman8:
## Summary
- After a real-project Firestore seed, automatically call `/api/revalidate` so a newly added context does not stay empty on wahl.chat for an hour (ISR still holding the pre-seed party list).
- The revalidate route now accepts `tags` / `paths` arrays in one request, and uses layout scope so a context root also busts nested pages.
- A cloud seed without `REVALIDATE_SECRET` still writes Firestore, then exits with a pointer to the Vercel env-var page instead of leaving the site stale.

## Test plan
- [ ] `REVALIDATE_SECRET=... ENV=prod python firebase/scripts/seed_firestore.py` prints a successful revalidation against https://wahl.chat/api/revalidate
- [ ] The same command without `REVALIDATE_SECRET` writes Firestore, then exits and prints https://vercel.com/wahl-chat/web/settings/environment-variables
- [ ] An emulator seed (`FIRESTORE_EMULATOR_HOST` set) skips revalidation
- [ ] `cd ai-backend && uv run python -m pytest tests/test_seed_revalidate.py`

### DIFF
--- a/firebase/scripts/revalidate_frontend_cache.py
+++ b/firebase/scripts/revalidate_frontend_cache.py
@@ -2,10 +2,11 @@
 # SPDX-FileCopyrightText: 2026 wahl.chat
 #
 # SPDX-License-Identifier: PolyForm-Noncommercial-1.0.0
-"""Bust the Next.js ISR cache for seeded Firestore data.
+"""Bust the Next.js data cache for seeded Firestore data.
 
-The live site caches contexts and parties for about an hour. After a
-real-project seed, call this so a newly added election does not stay empty.
+The hosted app caches election config (contexts, parties, sources,
+questions) by tag. After a real-project seed, call this so a newly added
+election does not stay empty.
 
 Usage (from the repo root or firebase/):
 
@@ -43,9 +44,9 @@ def _data_dir() -> Path:
     return FIREBASE_DIR / "firestore_data" / _env()
 
 
-# Tags the web app attaches to unstable_cache / ISR for the collections the
-# seed script writes.
-SEED_CACHE_TAGS = (
+# Coarse tags still attached on the web cache (covers entries written before
+# per-context tags). Also busted so a deploy mid-seed cannot leave both.
+COARSE_CACHE_TAGS = (
     "contexts",
     "context_parties",
     "proposed_questions",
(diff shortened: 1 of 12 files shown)


firebase/scripts/seed_firestore.py
```diff
@@ -347,6 +352,33 @@ def seed_sources(db):
...
+        print("\nSkipping frontend cache revalidation (Firestore emulator).")
+        return
+
+    try:
+        print()
+        frontend_cache.post_revalidate(context_ids)
+    except frontend_cache.RevalidateConfigError:
+        print()
+        print(frontend_cache.missing_secret_warning())
+    except frontend_cache.RevalidateRequestError as exc:
```
SimonKaran13, line 385: Could this make the seed exit non-zero? A wrong secret or failed request is swallowed here, so the script still prints “Seeding complete” and automation thinks the cache was flushed.

firebase/scripts/revalidate_frontend_cache.py
```diff
@@ -0,0 +1,198 @@
...
+def site_url() -> str | None:
+    """Public origin of the Next.js app whose cache we must bust.
+
+    REVALIDATE_URL / SITE_URL win; otherwise ENV=prod → wahl.chat and
+    ENV=dev → dev.wahl.chat.
+    """
+    explicit = os.getenv("REVALIDATE_URL") or os.getenv("SITE_URL")
+    if explicit:
+        return explicit.rstrip("/")
+    return DEFAULT_SITE_URLS.get(_env())
```
SimonKaran13, line 126: Can we validate the Firestore project against ENV before this? ADC can seed one project while this flushes another deployment.

SimonKaran13 (approved):
Check out the comments and then we can merge this

firebase/scripts/seed_firestore.py
romman8: I would not make cache busting errors fail the script as this can easily be repaired manually/self-heals after ~1h. We print a warning to indicate the cache was not busted to make it cleary visible to whoever invokes the script that this part was not successful --> leaving as is for now

firebase/scripts/revalidate_frontend_cache.py
romman8: Good point, will do

romman8: Done. `post_revalidate` now checks the Firestore client project against ENV (`wahl-chat` / `wahl-chat-dev`) before picking a site URL. A mismatch warns and skips the flush so ADC cannot bust the other deployment. The standalone script still ignores leftover `GOOGLE_CLOUD_PROJECT` from local gcloud / `.env`.

PR #158, opened 2026-08-30
Move chat to Gemini 3.6 Flash and Gemini 3.1 Flash-Lite

romman8:
## Summary

We move chat to newer language models. The current models give weaker answers. They add latency. Google will stop some of them. Cost increases only a little.

The main answer model is Gemini 3.6 Flash with minimal reasoning. The pre- and post-processing model is Gemini 3.1 Flash-Lite. If the first model fails, the next model in the list is used: GPT-5.6 Terra then older Gemini Flash models for answers, and GPT-5.6 Luna then older Flash-Lite models for pre- and post-processing. All of these models use temperature 1.

We remove Azure OpenAI. Each new OpenAI model would need a manual deploy on Azure. We do not want to maintain that.

We also remove LLM size, premium-only models, and the unused per-model rate-limit flag. Every user now uses the same model list. The web client no longer sends those flags.

## Test plan

- [ ] Run backend unit tests for the new model list and cache keys.
- [ ] Confirm Vertex Gemini models still run before AI Studio twins when Vertex credentials exist.
- [ ] Send a chat message and confirm a streamed answer.
- [ ] Confirm the voting-behavior summary still streams.
- [ ] Confirm the rate-limit banner still appears only when all models fail.

### DIFF
--- a/ai-backend/tests/test_llms_vertex.py
+++ b/ai-backend/tests/test_llms_vertex.py
@@ -49,17 +49,6 @@ def test_no_vertex_tier_without_credentials() -> None:
     ]
 
 
-def test_no_azure_clients() -> None:
-    """Azure OpenAI is not registered; new OpenAI models would have to be
-    deployed there by hand, which we are not maintaining.
-    """
-    names = [
-        entry.name
-        for entry in llms.RESPONSE_GENERATION_LLMS + llms.PRE_AND_POST_PROCESSING_LLMS
-    ]
-    assert not any(name.startswith("azure-") for name in names)
-
-
 def _ai_studio_gemini_clients():
     return [
         entry.model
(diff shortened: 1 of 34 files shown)


ai-backend/tests/test_llms_vertex.py
```diff
@@ -31,9 +31,41 @@ def test_no_vertex_tier_without_credentials() -> None:
...
+
+def test_no_azure_clients() -> None:
+    """Azure OpenAI is not registered; new OpenAI models would have to be
+    deployed there by hand, which we are not maintaining.
+    """
+    names = [
+        entry.name
+        for entry in llms.RESPONSE_GENERATION_LLMS + llms.PRE_AND_POST_PROCESSING_LLMS
+    ]
+    assert not any(name.startswith("azure-") for name in names)
```
romman8: remove this fragile test that only asserts that a removal done in this PR and carries no relevance for future changes

PR #163, opened 2026-09-06
feat(seo): serve a real landing page at / as one full-screen panel

RobinFrasch:
`/` was a 307 to whichever election geo-IP picked, so it never returned content. Every inbound backlink — tum.de, cit.tum.de, kas.de, politikzumanfassen.de, go-vote.de — landed on a redirect, and the destination moved every time an election concluded, so authority never accumulated anywhere. Googlebot crawls from US IPs, so it was also sent somewhere different from a German visitor.

`/` now renders the same page for everyone. The featured election is derived from the context dates via `lib/elections.ts`, so nothing varies by requester and nothing goes stale as elections conclude. Verified that a Googlebot request from a US IP and a plain German request receive byte-identical `<head>` and visible text.

## The page

image

One panel, exactly the height of the viewport — not the election page's furniture under a hero band. `/` moves out of the `(with-header)` shell into its own route group, so there is no header or footer bar competing with it:

- the wordmark on top;
- **h1** — "Mit wahl.chat Parteipositionen zur Wahl vergleichen: im Chat mit Quellen aus offiziellen Dokumenten.";
- a subheadline naming the evidence: Plenarprotokolle, Bundestag recordings, namentliche Abstimmungen, Wahlprogramme;
- **one call to action** — a large button into the featured election, its date beneath it, and a quiet "Andere Wahl wählen" that opens the full picker.

Anleitung, Über uns, Impressum and Datenschutz move into the panel. With no footer they would otherwise be unreachable from `/`, and German law requires the last two to stay one click from every page.

`election-select-items.tsx` carries the option rendering for the election pages' selector, replacing two verbatim copies of the same block inside it.

### Internal linking

All seven elections are real anchors in the server HTML, with the full election name as the anchor text — so `/` passes the authority from its inbound links on to each election page, and they are discovered by crawl rather than only by sitemap. The picker this replaces was a listbox of router pushes, which a crawler cannot follow.

The featured election is the call to action. The other upcoming ones sit under "Andere Wahl auswählen" as small bordered buttons. Concluded elections sit behind a native `<details>`, collapsed: there is no live search demand for an election that is over, and a disclosure element keeps the links in the markup, where a conditionally rendered toggle would not.

An `ItemList` node mirrors the same set in the same order, so the structured data and the markup cannot drift apart.

## Geo routing is kept — just not on `/`

`REGION_CONTEXTS` and its five-day past-election buffer are unchanged. What changes is which paths consult them:

- **`/` no longer does.** It falls through to the page, and stays in the matcher only for the tenant header and the budget kill switch.
- **The legacy context-less links now do** — `/chat/{party}`, `/session/*`, `/swiper*`, `/share`, `/sources` — instead of the hardcoded `DEFAULT_CONTEXT_ID`. That constant names Sachsen-Anhalt, whose election was 6 September, so within days every one of those links would have pointed at a concluded election; the region lookup already skips those.

These paths are redirects either way, are never listed in `sitemap.xml`, and are not canonical URLs, so letting them vary by region costs nothing. The defect was specifically that `/` — the backlink target — varied.

`middleware.test.mjs` keeps its region loop, retargeted onto `/sources`, and gains a case asserting `/` is never redirected whatever the request claims about its origin.


web/app/(landing)/page.tsx
```diff
@@ -0,0 +1,174 @@
...
+// The featured election is derived from the context dates, so it rolls over on
+// its own once the current one concludes, and the output does not depend on the
+// requester — crawlers and visitors get the same page, which the geo redirect
+// this replaced could not promise. Freshness comes from the context read, which
+// is cached and busted by tag on seed; a page-level revalidate would pin an
+// empty list into a shell instead.
+export const dynamic = 'force-dynamic';
+
+const TITLE =
+  'wahl.chat – Parteipositionen zur Wahl im Chat vergleichen, mit Quellen';
```
romman8: Is this really the best tagline for SEO and humans?
"Parteipositionen" does not sound great.
Maybe sth like:
A "Wahlprogramme vergleichen: Informiert wählen mit wahl.chat",
B "Parteien & Wahlprogramme vergleichen mti wahl.chat"

I read that:
1. keywords should be fron-loaded
2. brand should be at the end
3. we need it to be under 60 characters

web/app/(landing)/page.tsx
```diff
@@ -0,0 +1,174 @@
...
+const DESCRIPTION =
```
romman8, line 53: Let's also improve this. Should be between 140 and 155 characters and use an active voice + have a CTA.
Option A:
"Welche Partei passt zu dir? Vergleiche Wahlprogramme zu Bundestags-, Landtags- und Kommunalwahlen im KI-Chat. Stelle direkt Fragen an die Parteien und wähle perfekt informiert. Jetzt chatten!"
Option B:
"Die interaktive Ergänzung zum Wahl-O-Mat. Nutze modernste KI belegt durch Plenarprotokolle, namentliche Abstimmungen und Wahlprogramme, um Parteien wirklich zu verstehen. Jetzt ausprobieren und informiert wählen!"

web/app/(landing)/page.tsx
```diff
@@ -0,0 +1,174 @@
...
+      <section className="relative flex min-h-dvh w-full flex-col overflow-hidden">
+        <BrandBlurBackdrop />
+
+        <div className="relative flex flex-1 flex-col items-center justify-center gap-5 px-5 py-8 text-center md:gap-6 md:py-14">
+          <Logo variant="large" className="h-8 w-auto md:h-10" />
+
+          <div className="flex max-w-3xl flex-col gap-4">
+            <h1 className="text-balance text-2xl font-bold tracking-tight text-foreground sm:text-3xl md:text-4xl">
+              Mit wahl.chat Parteipositionen zur Wahl vergleichen: im Chat mit
+              Quellen aus offiziellen Dokumenten.
```
romman8: Also not a fan of this copy.
Maybe something like:
"Eigene Frage stellen, Parteipositionen verstehen, informiert wählen." (could be "Eigene Frage stellen. Parteipositionen verstehen. Informiert wählen." if we want to be a bit more modern)
Open for suggestions though

web/app/(landing)/page.tsx
```diff
@@ -0,0 +1,174 @@
...
+            </h1>
+
+            <p className="text-pretty text-sm text-muted-foreground sm:text-base md:text-lg">
```
romman8: For SEO, this should probably be `<h2>` adjusted to look less heavy with css, no?

web/app/(landing)/page.tsx
```diff
@@ -0,0 +1,174 @@
...
+              Individuelle Fragen, individuelle Antworten, belegt durch
+              Plenarprotokolle, Mitschnitte aus dem Bundestag, namentliche
+              Abstimmungen und Wahlprogramme.
```
romman8: Suggestion:
"Welche Partei passt zu dir? Vergleiche Parteien belegt durch Plenarprotokolle, namentliche Abstimmungen und Wahlprogramme"

romman8 (changes requested):
The new landing page is definitely a step in the right direction.

Let's definitely iterate on copy though

Additionally,  it feels super odd that I cannot scroll on the page. You can check https://knowunity.de/ and https://lovable.dev/de/ for stuff that one could put below (media links, wahl.chat stats, background, FAQ, ...)

Would btw also be an excellent follow-up to have links to content pages with many keywords on the landing page (e.g. we could host summaries of the Plenarprotokolle, Abstimmungen, or sample wahl.chat responses/conversation snippets on subpages and link to them See https://machtblick.de/ for inspiration.)

Image

web/app/(landing)/page.tsx
RobinFrasch: Some issues:
- Wahlprogramme gives the wrong, limited impression
- Comparing parties is not ideal because I want to avoid any phrasing that implies that we speak for the parties

Suggestion:
- "Parteipositionen mit Quellen vergleichen - wahl.chat"
- Politik in einfach: Dein KI Chat vor der Wahl

Let's discuss

web/app/(landing)/page.tsx
RobinFrasch: Suggestion:
- Dein verlässlicher Zugang zu Wahlprogrammen, Plenarprotokollen und Abstimmungen vor den kommenden Wahlen. Jetzt chatten und informiert wählen!

web/app/(landing)/page.tsx
RobinFrasch: agree, but I tried to work keywords like "Wahl", "Parteiposition", "Chat" and so on in.
Maybe:
- Mit wahl.chat vorbereitet in die nächste Wahl: Deine Fragen, tausende Quellen zu Parteipositionen und eine verlässliche Antwort.

web/app/(landing)/page.tsx
RobinFrasch: Claude says no

web/app/(landing)/page.tsx
RobinFrasch: Dies gute
Fits a less detailed but more expression of how much data we make accessible on the top line

web/app/(landing)/page.tsx
RobinFrasch: Settled on: Mit wahl.chat Politik verstehen - dein KI Chat vor der Wahl

web/app/(landing)/page.tsx
RobinFrasch: Settled on: Die KI-basierte Ergänzung zum Wahl-O-Mat: Verstehe Parteien anhand von Plenarprotokollen, Abstimmungen und Wahlprogrammen. Jetzt informiert wählen!

web/app/(landing)/page.tsx
RobinFrasch: Settled on: Fragen stellen. Politik verstehen. Informiert wählen.

web/app/(landing)/page.tsx
RobinFrasch: Settled on: Welche Partei passt zu dir? Vergleiche Parteien belegt durch Plenarprotokolle, namentliche Abstimmungen und Wahlprogramme

Switch to h2 and style small to look similar to paragraph right now in grey

RobinFrasch:
TODO:
- sick animation for logo
- move CTA to top once scrolled past
- have icon of states shown behind example questions for elections

PR #164, opened 2026-09-06
fix: unbreak the BUDGET_SPENT kill switch and correct the Context.date type

RobinFrasch:
- `BUDGET_SPENT` was checked before the static-page check, and `/budget-spent` is itself a static page — the kill switch redirected its own landing page to itself forever, and sent `/sitemap.xml`, `/robots.txt` and `/manifest.json` there too.
- `Context.date` was declared `string | null` but `firebase-server.ts` maps it through `firestoreTimestampToDate()`, so it is a `Date` at runtime; `formatGermanDate` only worked on it by accident. It now accepts both, and its `if (!date)` guard is replaced with a real Invalid-Date check — `new Date()` never returns null for a bad input.
- `AiDisclaimer` accepted a `className` and never applied it. Every page renders it twice, once gated to desktop and once to mobile, so both copies showed at every width and the notice appeared under the input twice. Wording is unchanged.
- `/chat` and `/session` now 308 to `/`, which is a stable target; paths carrying a party or session id still need a context and stay 307. `startsWith('/chat')` became an exact match plus `/chat/`, so `/chatfoo` is no longer parsed as a party link.
- Adds `/sitemap.xml`, `/robots.txt`, `/manifest.json` and the icon routes to the static-page check, since the `/:contextId` matcher was catching them.
- Deletes `app/(with-header)/swiper/page.tsx` — unreachable since middleware intercepts `/swiper`, still shipping a 407 kB bundle.
- Drops the unconditional 500 ms `setTimeout` that forced skeleton party cards even though the server layout had already supplied the parties.
- Adds unit tests for the election date rule (five-day buffer, undated and unparseable dates, all-concluded list) and extends the middleware tests to cover the kill switch, the metadata routes and the 308s. Uses the Bun runner, so no test framework is pulled in.


web/components/legal/ai-disclaimer.tsx
```diff
@@ -100,7 +100,15 @@ type Props = {
 function AiDisclaimer({ className }: Props) {
   return (
     <ResponsiveDialog>
-      <p className="my-2 text-center text-xs text-muted-foreground">
+      {/* The caller decides which breakpoint this belongs to — the page renders
+          one instance for desktop and one for mobile, so dropping className
+          here shows both at once. */}
```
romman8: can remove the slop comment

romman8 (approved):

PR #168, opened 2026-09-07
Feat/pledgetracker

miquelcamprubi:
## Why

wahl.chat can say what a party promises. It cannot say what happened after the
promise. This PR adds that missing half.

PledgeTracker is a University of Cambridge research project ([EMNLP 2025 demo
paper](https://aclanthology.org/2025.emnlp-demos.64/)), built with the TUM
Social AI group. It collects dated parliamentary evidence for a pledge. This PR
shows that evidence in the chat answer, with a link to every source.

Cambridge labels each event `Ja` or `Nein`. The label means "useful evidence for
tracking". It does not mean "kept" or "broken". The UI therefore says "Ziele",
never "Versprechen", and no text claims that a party kept or broke a pledge.

## What it does

A party answers a question. If the topic matches a tracked pledge of that party,
a pledge card appears below the answer. A click opens a popup with up to three
pledges. Each pledge has a timeline of dated, source-linked evidence. If nothing
matches, no card appears.

## Screenshots

Layout only: the pledge and its timeline below are mock data.

|  | Mobile | Desktop |
| --- | --- | --- |
| Card in the answer |  |  |
| Popup |  |  |

## How it works

**Firestore stores the pledge. Qdrant only finds it.** Firestore holds the full
record at `pledges/{pledge_id}`, including the timeline. Qdrant holds one small
vector per pledge, under a new `pledge_record` source type. Timelines are never
embedded. No client reads the collection, because the browser gets pledges over
the chat stream, so `firestore.rules` needs no new rule.

The pledge id is `uuid5(party:claim:region:pledge_date)`, so re-ingestion upserts
in place. A reconcile step retires pledges whose registry row changed or
disappeared. Region codes are ISO 3166-2, to match the context seeds.

**Ingestion (`connectors/pledgetracker/`).** The Cambridge API is a job queue in
front of one GPU: one job at a time, about 3 minutes per pledge. We own the
pledge list (`data/pledges/*.jsonl`). The runner is therefore incremental: a
freshness watermark, bounded batches for the 15-minute job cap, resume from the
stored job id, bounded retries, and least-recently-checked pledges first.
Reconcile refuses to retire more pledges than the registry holds, unless you
pass `--force-reconcile`. The runner stays out of `CONNECTOR_FACTORIES`, because
pledges have no monotonic cursor and it writes to two stores.

**Chat lookup.** The answer text streams first. The backend then reuses the
answer's RAG query and embedding to search Qdrant. An LLM gate drops off-topic
candidates, Firestore hydrates the rest, and the result rides on the existing
`party_complete` event.

The lookup is awaited before `party_complete`, so it delays that final frame. It
does not delay the answer text, which is already complete. The 8-second gate
timeout bounds the worst case. Moving the lookup off the answer path is a
follow-up.

Failures are safe. A gate failure keeps all candidates, a lookup failure returns
nothing, and the answer stays untouched. Comparison turns and the wahl.chat
meta-party are skipped.

**Frontend.** No new dependencies. The card is the trigger, and the popup reuses
`ResponsiveDialog`. Source links accept http and https only. A green glow marks
the newest answer, and stops once the user opens the popup in that session. The
payload persists with the session, so pledges survive a reload. Three cookieless
Vercel events record the offer, real viewport exposure, and the click.

## Config and ops

- `PLEDGETRACKER_API_KEY` and `PLEDGETRACKER_ENABLE_LIVE` enable live runs. The
  key lives only in the gitignored `.env`. Without them the runner uses an
  offline fixture.
- Run `make run-pledgetracker`. AGENTS.md holds the runbook and the data rules.
- Two registries ship: a one-pledge federal sample, and 15 Sachsen-Anhalt
  pledges. They share no region, so a default run cannot retire real pledges.
  The existing `landtagswahl-sachsen-anhalt-2026` context already covers them.

## Testing

48 new backend tests cover the queue client, the registry and id rules, the
mapper, batch order, reconcile scope and its guard, the remote-target guard, and
the gate fallbacks. All network, LLM and store access is faked.

Verified locally on `ef95fbb`. Backend: 971 passed. The 5
`test_seed_revalidate` failures are pre-existing test-order dependence: that
file passes 14/14 on its own, and this branch changes no seed or revalidate
file. SSE smoke: 12 passed. `ruff`, `mypy` and the GDPR wall guard are clean.
Web `lint`, `typecheck` and `build` are clean, and `bun test` passes 22.

A live run against the real queue API was validated on the TUM research fork that
this port derives from. The offline demo:

```bash
make stores-up && make bootstrap-collection
FIRESTORE_EMULATOR_HOST=localhost:8081 make seed-local
FIRESTORE_EMULATOR_HOST=localhost:8081 make run-pledgetracker   # stub fixture
make dev   # any context → SPD → ask about Mindestlohn
```

## Out of scope

Cloud Run Job wiring for scheduled ingestion. Moving the chat lookup off the
answer path. Coalition pledges go to the lead party only; fan-out to all
partners is a follow-up. `score_threshold` tuning before the gate, once the
pledge corpus grows.


ai-backend/data/pledges/sample_pledges.jsonl
```diff
@@ -0,0 +1,2 @@
+{"claim": "Wir wollen, dass die A14 und die A143 spätestens 2025 komplett fertig gestellt sind", "pledge_date": "2021-03-27", "pledge_author": "CDU", "party": "CDU", "bundesland": "Sachsen-Anhalt", "time_range": "since_pledge_date", "policy_area": "Verkehr"}
+{"pledge_id": "spd-mindestlohn-12-euro-de-2021", "claim": "Wir werden den gesetzlichen Mindestlohn in einer einmaligen Anpassung auf 12 Euro pro Stunde erhöhen.", "pledge_date": "2021-11-24", "pledge_author": "SPD", "party": "SPD", "time_range": "since_pledge_date", "normalized_summary": "Einmalige Erhöhung des gesetzlichen Mindestlohns auf 12 Euro pro Stunde.", "policy_area": "Arbeit und Soziales", "pledge_source_title": "Koalitionsvertrag 2021–2025: Mehr Fortschritt wagen", "pledge_source_url": "https://www.spd.de/fileadmin/Dokumente/Koalitionsvertrag/Koalitionsvertrag_2021-2025.pdf", "pledge_source_locator": "Mindestlohn-Abschnitt, ca. S. 54", "context_id": "bundestagswahl-2025"}
```
romman8, line 1: What's the semantic difference between "pledge_author" and "party"? How is "pledge_author" used downstream? What are the reasons we can't just use the "party" for that?

miquelcamprubi: Both fields come from the Cambridge queue API's job schema rather than from us: pledge_author is the required, verbatim author attribution of the pledge (it can be a coalition or a state association, e.g. "CDU Sachsen-Anhalt, SPD Sachsen-Anhalt, FDP Sachsen-Anhalt"), while party is an optional filter their evidence-retrieval pipeline applies when searching parliamentary documents. Our client defaults party to pledge_author when it's omitted ([registry.py](http://registry.py/)::job_inputs()) so in these sample rows it's admittedly redundant. It's spelled out to mirror the upstream schema 1:1, so a future full pledge dump from Cambridge maps onto PledgeInput without surprises.

Downstream on our side, neither field is the wahl.chat attribution: that's party_id (canonical slug), derived from pledge_author via the shared party_slugs tables, with an explicit party_id override for coalition pledges (lead-party convention). The verbatim pledge_author is kept on the Firestore record for provenance. We can drop the redundant party values from the sample rows if you'd prefer the file minimal.

ai-backend/src/ingestion/connectors/pledgetracker/bulk.py
```diff
@@ -0,0 +1,492 @@
...
+    chunks_upserted: int
+
+
+def _guard_firestore_target() -> None:
+    """Prevent accidental dev/local writes to real Firestore."""
+    env = os.getenv("ENV", "dev")
+    if env != "prod" and not os.getenv("FIRESTORE_EMULATOR_HOST"):
+        raise RuntimeError(
+            "FIRESTORE_EMULATOR_HOST is required for PledgeTracker ingestion "
+            "when ENV is not prod. Set it to localhost:8081 for local runs."
```
romman8: What is the expected behavior for `env=="dev"`? Imo we should also be able to ingest pledges into our dev environment

ai-backend/src/pledge_tracker_service.py
```diff
@@ -0,0 +1,161 @@
...
+    *,
+    query: str,
+    party_id: str,
+    context_id: str,
+    query_vector: Optional[list[float]] = None,
+    limit: int = 3,
+) -> Optional[PledgeTrackerSuggestions]:
+    """Retrieve gated pledge suggestions for one party answer, or None."""
+    context = await aget_context_by_id(context_id)
+    region_path = context.region_path if context and context.region_path else ["DE"]
```
romman8: Small perf nit: this re-fetches the context just to get `region_path`, but the caller already has it. `generate_chat_stream` deliberately fetches it once up front (`chat_service.py:1678`) — the comment there says "Fetch context ONCE; derive region_path ... All retrieve() calls in both paths reuse these" — and `fetch_party_response_stream` already receives `region_path` as a parameter.

Could we take `region_path` as an argument here and drop the lookup? Saves a Firestore read per party per answer, and it guarantees the pledge filter uses the same region scope as the rest of the retrieval instead of a second, independently-derived one. The cached-replay path would need it threaded in too.

ai-backend/src/ingestion/connectors/pledgetracker/bulk.py
```diff
@@ -0,0 +1,492 @@
...
+            # Best-effort compact headlines for the UI; failures leave
+            # event_short=None and the frontend falls back to full text.
+            add_short_titles(record)
+
+            _write_pledge_record(db, record)
+            doc_ref.set({"pledgetracker_last_error": None}, merge=True)
+
+            chunk = connector.to_chunk(record)
+            vectors = _embed_texts(embed, [chunk.text])
+            _upsert_chunks(qdrant, collection_name, [chunk], vectors)
```
romman8: The Firestore write runs before the Qdrant write, and `merge_result` sets `last_checked_at` — so line 273 marks the pledge as fresh.

If the run stops between line 273 and line 278, the pledge has no vector. The realistic triggers are the 15-minute job cap and an OpenAI outage that uses up the retries in `_embed_texts`. `_is_fresh` then skips the pledge for the next 7 days. `_resumable_job_id` does not recover it either, because `last_checked_at` is now newer than `pledgetracker_job_submitted_at`. The pledge stays out of retrieval until the watermark expires, and the repair costs a new ~3-minute GPU run.

Can we write Qdrant first? That order is safe in the other direction: `aget_pledges_by_ids` already skips a retrieval hit with no document and calls it a stale point, not an error. If the order has to stay, writing `last_checked_at` in its own merge after line 278 also closes the window.

ai-backend/src/ingestion/connectors/pledgetracker/bulk.py
```diff
@@ -0,0 +1,492 @@
...
+        except Exception as exc:  # noqa: BLE001
+            print(
+                f"WARNING: pledge {pledge_id} (job {job_id}) failed: {exc}",
+                file=sys.stderr,
+            )
+            doc_ref.set({"pledgetracker_last_error": str(exc)}, merge=True)
```
romman8: This path writes only the error. It never writes `last_checked_at`, so the pledge stays stale forever. The batch takes pledges in registry order (lines 197-208), so a permanently failing pledge keeps one of the three slots on every run. Three bad rows at the top of a JSONL and no other pledge is ever processed.

There is also a case that repeats without end. A job with status `done` and no `events` list makes `merge_result` raise. The Firestore write never happens, so the next run resumes the same job id, gets the same payload, and fails again. There is no attempt counter and no backoff.

Can we write a `last_attempted_at` here, and order the batch by oldest check first? That bounds the retries and lets the rest of the registry through.

ai-backend/src/ingestion/connectors/pledgetracker/registry.py
```diff
@@ -0,0 +1,205 @@
...
+        same Firestore doc and Qdrant point instead of duplicating them.
+        """
+        if self.pledge_id:
+            return self.pledge_id
+        return str(
+            compute_source_item_id(
+                SourceType.PLEDGE_RECORD.value,
+                f"{self.resolved_party_id()}:{self.claim}:{self.region}",
+            )
+        )
```
romman8, line 166: The id includes the claim text, so any edit to a claim makes a new id. The next run then writes a second Firestore document and a second Qdrant point, and never touches the old pair. Removing a row from the JSONL leaves both stores unchanged too. In both cases the old pledge stays publicly readable and stays in retrieval results.

`manifestos/supersede.py` handles this for the manifesto side, and AGENTS.md states the rule for `manifesto_uploads`: removing a manifest line retires that document's chunks on the next run. This runner has no equivalent, and it also skips the orphan check in `run.py` because it calls `_upsert_chunks` directly.

Could we add a reconcile step after the batch — scroll `source_type=pledge_record` scoped to the registry's own `region` / `context_id` set, then delete the points and documents whose `pledge_id` is not in the registry? The scoping keeps one registry from deleting another registry's pledges.

One related note: `pledge_date` is not part of the id. Two rows with the same party, claim and region but different dates collapse into one record, and the later run overwrites the earlier `pledge_date` and `pledge_source_title`. Rows 5 and 13 of `sachsen_anhalt_pledges.jsonl` are both the 103 % Unterrichtsversorgung pledge from different years, and only the wording keeps them apart today.

web/components/chat/chat-pledge-tracker.tsx
```diff
@@ -0,0 +1,345 @@
...
+            className="underline-offset-2 hover:text-foreground hover:underline focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
+          >
+            {showFullText ? 'Weniger' : 'Mehr'}
+          </button>
+        )}
+        {event.url && (
+          <a
+            href={event.url}
+            target="_blank"
+            rel="noreferrer"
```
romman8: `event.url` comes straight from the queue API and reaches `href` without a scheme check. `connector.py:275` copies it with `url=raw.get("url")`, the model types it as a bare `Optional[str]`, and this component does not validate it. A `javascript:` or `data:text/html,` value runs in the user's session, with their Firebase auth in scope. `pledge.pledge_source_url` on line 152 has the same gap.

Can we accept `http:` and `https:` only? A small helper in `lib/pledge-tracker/pledges.ts` that returns the value or `null` would let the link fall back to plain text, and a check in `normalize()` would keep bad values out of Firestore as well.

Minor, same lines: the other external links in the repo use `rel="noopener noreferrer"`. `noreferrer` implies `noopener` in current browsers, so this is consistency only.

romman8 (commented):
Four things in the ingestion runner and the popup that I think need a change before this goes in. None of them affect the feature design, which reads well — they are about what happens when a run stops early, when a pledge fails, when a claim gets edited, and what the API is allowed to put into an `href`.

web/components/chat/chat-pledge-tracker-button.tsx
```diff
@@ -0,0 +1,106 @@
...
+  // The button only renders when pledge suggestions exist, so mounting IS the
+  // impression (exposure) — tracked once per message.
+  useEffect(() => {
+    if (trackedImpressions.has(message.id)) return;
+    trackedImpressions.add(message.id);
+    track('pledge_tracker_impression', {
+      party: partyId,
+      message: message.content ?? 'empty-message',
+    });
+  }, [message.id, message.content, partyId]);
```
romman8: `mount = exposure` does not quite hold here. `trackedImpressions` is module level, so a reload sends a fresh event for every rehydrated message. And `ChatGroupedMessages` renders all carousel slides into the DOM at once, so in a group chat every party's button fires while only one slide is visible. Messages scrolled far off-screen fire too. A CTR computed against this denominator comes out understated, and it is not comparable between users, because heavy reloaders dilute it.

I think the event is being asked two different questions. Could we split it?

- `pledge_tracker_offered` — fire once in `completeStreamingMessage` when the payload arrives. This answers "how often does the feature have something to show". It is exact, needs no DOM, needs no dedupe, and never fires for history.
- `pledge_tracker_impression` — real exposure only, so it can serve as the CTR denominator.

For the exposure half, one `IntersectionObserver` on the existing wrapper `div` covers both cases at once: `CarouselContent` wraps the slides in `overflow-hidden`, so an inactive slide reports ratio `0` exactly like an off-screen message — no carousel-specific code needed. Fire at `>= 0.5` ratio held for ~1s so a fast scroll does not count, disconnect after, and keep the dedupe in `sessionStorage` so a reload does not re-fire. `sticky-input.tsx:27` already uses the raw API, so this adds no dependency.

Two small things while in here:

- Claim the impression in `handleClick` too. Otherwise a click inside the dwell window gives a click with no impression, and CTR goes above 100 %.
- `message: message.content` sends the full answer body on every fire. `message_id` plus the pledge count would be enough, and it keeps the answer text out of the analytics store.

Separately: `pledge_tracker_modal_viewed` looks redundant. The button is the only way to set `open=true`, so it is 1:1 with `pledge_tracker_button_clicked`.

ai-backend/src/ingestion/connectors/pledgetracker/bulk.py
miquelcamprubi: Good point. The guard was blocking legitimate dev-environment ingestion. Added --allow-remote as an explicit opt-in: still refuses by default outside prod, and warns which ENV it's writing to. Fixed in f1a3be4.

ai-backend/src/pledge_tracker_service.py
miquelcamprubi: You're right, that was a wasted read. The service now takes region_path from the caller (both the live and the cached-replay path thread through the value generate_chat_stream fetched once) so the pledge filter is guaranteed to share the stream's region scope. Fixed in 03e573a.

ai-backend/src/ingestion/connectors/pledgetracker/bulk.py
miquelcamprubi: Swapped the order: Qdrant first, then the Firestore write that carries the watermark. A crash between the two now just means the pledge is redone next run. Fixed in aa34c00.

ai-backend/src/ingestion/connectors/pledgetracker/bulk.py
miquelcamprubi: Right on both counts. Failures now stamp pledgetracker_last_attempted_at, and the batch takes the least-recently-touched pledges first (never-touched before everything), so failing rows rotate to the back instead of blocking the registry. Job ids are kept, so transient failures still resume the stored server-side result for free. Fixed in 8b38575.

ai-backend/src/ingestion/connectors/pledgetracker/registry.py
miquelcamprubi: Pledge_date is now part of the identity seed. The same wording pledged in two election cycles is two pledges (your rows 5/13 example); changed now while prod holds no pledges, so no migration needed. And live runs end with a reconcile that retires store pledges absent from the registry, scoped to the registry's own regions. Fixed in 516cbc7.

web/components/chat/chat-pledge-tracker.tsx
miquelcamprubi: You're right, that was a real hole. URLs are now gated twice: the connector drops non-http(s) values before they reach Firestore, and the frontend validates again and falls back to plain text. External links also carry rel="noopener noreferrer" now. Fixed in cdc840f.

web/components/chat/chat-pledge-tracker-button.tsx
miquelcamprubi: Agreed with all of it. Mount really was answering two questions. Implemented your split: pledge_tracker_offered fires once per live party_complete (exact, DOM-free, never for history), and pledge_tracker_impression is real exposure via IntersectionObserver. Inactive carousel slides report ratio 0 through the overflow clipping, exactly as you said. Events carry message_id + pledge count instead of the answer text, and modal_viewed is gone. Fixed in 94e5b4c.

Makefile
```diff
@@ -266,6 +266,18 @@ run-manifestos: bootstrap-collection
 	cd ai-backend && $(QDRANT_ENV) \
 		uv run python -m src.ingestion.connectors.manifestos.bulk $(ARGS)
 
+# PledgeTracker: stub mode ingests the packaged fixture. Live mode (set
+# PLEDGETRACKER_ENABLE_LIVE=true + PLEDGETRACKER_API_KEY in ai-backend/.env)
+# submits registry pledges to the Cambridge queue API (~3 min GPU time per
+# pledge, one at a time). Requires FIRESTORE_EMULATOR_HOST unless ENV=prod.
```
romman8: "Requires FIRESTORE_EMULATOR_HOST unless ENV=prod."
Why? Should also work against the dev Firebase environment

romman8: Real gap, not just stale docs. `--allow-remote` already existed for deployed-dev Firestore, but local `.env` and the Makefile always set `FIRESTORE_EMULATOR_HOST`, so the guard treated that as emulator mode and the flag was a no-op.

`--allow-remote` now unsets the emulator host before `firebase_service` initializes, so it actually writes to the hosted ENV project (dev by default). Makefile comment updated.

`make run-pledgetracker ARGS="--allow-remote --registry data/pledges/sachsen_anhalt_pledges.jsonl"`

romman8 (approved):

PR #172, opened 2026-09-13
feat: record visible dwell time as first-party page visits

### LINKED ISSUE #171: Track visible dwell time in first-party page visits
## Why
The PledgeTracker Cambridge study needs per-user time-on-site (ms) and a way to tell whether someone returned for another session. Vercel Web Analytics cannot provide either: it has no session duration, and visitor hashes reset after 24 hours.

Chat sessions are also the wrong object. We only write `chat_sessions` on the first message, so bounce visits and return visits without a new conversation are invisible.

## What
Record **visible dwell time** globally as a first-party `page_visits/{visit_id}` document, keyed by the existing anonymous Auth `user_id`.

- Accumulate time the tab is visible (not wall-clock including background tabs).
- Persist `visit_id` in `sessionStorage` so a reload in the same tab stays one visit.
- Heartbeat every 20s plus flush on `visibilitychange` / `pagehide`.
- Stamp `visit_id` onto new `chat_sessions` so study metrics can be joined.
- Multiple `page_visits` for the same `user_id` is a return visit.

This is behavioral telemetry (paths + timing + ids), not political opinions, and must stay off `users/` and out of the ingestion/corpus path.

romman8:
Closes #171

## Summary
- Record visible tab time (not background-tab wall clock) in a new `page_visits` collection for every anonymous session, with a 20s heartbeat plus `pagehide` / keepalive flush.
- Stamp `visit_id` onto new chat sessions so PledgeTracker study metrics can be joined; multiple visits for the same `user_id` are return visits.
- Vercel Analytics cannot provide per-user duration or cross-day returns; this keeps the data first-party and off `users/`.

## Test plan
- [ ] Open the app locally with the Firestore emulator, confirm a `page_visits/{visit_id}` doc appears after anonymous auth.
- [ ] Background the tab and return: `visible_ms` should not jump by the hidden interval.
- [ ] Reload the same tab: same `visit_id`, `visible_ms` continues.
- [ ] Start a chat: the new `chat_sessions` doc has `visit_id`, and the visit doc lists that session id.
- [ ] Close the tab: last flush lands (client write and/or `POST /api/page-visit`).
- [ ] `cd web && bun test lib/page-visit`


romman8 (commented):
Nice split between the pure time accumulator and the transport, and the tests read well. `bun test lib/page-visit` and `bun run lint` are green locally; the `tsc` errors I see are pre-existing. Nothing here touches `users/`, so the Art. 9 wall stays intact.

Almost all of my comments are in the transport and authorization layer, not in the measurement logic. Four I would fix before merge:

1. `embedded` is always `false` in the browser.
2. The `sendBeacon` fallback can never authenticate.
3. The last flush on `pagehide` can be lost.
4. A crashed tab can lock a visit out of Firestore through the monotonic rule.

I could not test the rules, because this branch has no rules test harness. Please walk through the reload and crash cases against the emulator.

One point outside the code: this adds a first-party pseudonymous record plus a new `sessionStorage` entry. The privacy policy already covers Reichweitenmessung including Verweildauer, but a short legal check against § 25 TDDDG would be good before this ships.

web/components/providers/page-visit-provider.tsx
```diff
@@ -0,0 +1,117 @@
...
+      } catch {
+        // Best-effort: the client Firestore flush still works without a token.
+      }
+    };
+
+    const flushContext = () => ({
+      userId: user.uid,
+      pathname: window.location.pathname || pathnameRef.current,
+      tenantId: tenant?.id,
+      embedded: IS_EMBEDDED,
```
romman8: `IS_EMBEDDED` reads `process.env.IS_EMBEDDED`, which has no `NEXT_PUBLIC_` prefix. In a client component that value is always `undefined`, so `embedded` is never `true` — not even on the embed deployment. All other users of this flag are server components, which is why nobody hit this before.

Please pass the flag down from the server layout as a prop, or detect it in the browser with `window.self !== window.top`.

web/lib/page-visit/page-visit-request.ts
```diff
@@ -0,0 +1,62 @@
...
+    data: {
+      visitId,
+      visibleMs: clamped,
+      lastPath: optionalString(record.last_path),
+      landingPath: optionalString(record.landing_path),
+      startedAtMs,
+      contextId: optionalString(record.context_id),
+      tenantId: optionalString(record.tenant_id),
+      embedded: record.embedded === true,
+      idToken: optionalString(record.id_token),
```
romman8: `optionalString` drops every string longer than 512 characters. A Firebase ID token is a JWT of about 900–1200 characters, so `id_token` is always removed here and the route answers 401. The `sendBeacon` fallback can therefore never work.

Please give the token its own, larger length limit instead of reusing the path/id validator.

web/lib/page-visit/page-visit-flush.ts
```diff
@@ -0,0 +1,137 @@
...
+  beacon: boolean,
+): Promise<void> {
+  const visibleMs = currentVisibleMs();
+  try {
+    await upsertPageVisit(buildPayload(ctx, visibleMs));
+    markPageVisitCreated();
+  } catch (error) {
+    console.error('Failed to upsert page visit', error);
+  }
+  if (beacon) {
```
romman8: On `pagehide` we `await` the client Firestore write above. That write has no `keepalive`, so it usually does not finish during teardown, and this line may never run. The final dwell value is the one we care about most.

Please send the keepalive request (or `navigator.sendBeacon`) first, and treat the Firestore write as best effort after it.

web/lib/page-visit/page-visit.ts
```diff
@@ -0,0 +1,200 @@
...
+  if (!runtime || runtime.segmentStartedAt === null) {
+    return runtime?.snapshot.visibleMs ?? 0;
+  }
+  const elapsed = now - runtime.segmentStartedAt;
+  runtime.snapshot = {
+    ...runtime.snapshot,
+    visibleMs: clampVisibleMs(runtime.snapshot.visibleMs + elapsed),
+  };
+  runtime.segmentStartedAt = null;
+  persistPageVisitSnapshot(runtime.snapshot);
```
romman8, line 174: This is the only place that writes `visible_ms` to `sessionStorage`, but the heartbeat flushes `currentVisibleMs()` (stored value + open segment) to Firestore every 20 s.

If a visible tab crashes, the browser restores the old `sessionStorage` value, which is lower than the value already in Firestore. Every later client update then fails the `visible_ms >= resource.data.visible_ms` rule with permission-denied, and the visit silently stops updating.

Please persist the snapshot on each flush, or clamp the local value up to the last flushed value.

web/lib/firebase/firebase-admin.ts
```diff
@@ -29,6 +30,67 @@ import type { ShareableChatSessionSnapshot, Tenant } from './firebase.types';
...
+function getAdminApp(): FirebaseApp {
+  getDb();
+  return getApp();
+}
+
+export async function verifyFirebaseIdToken(idToken: string) {
+  return getAuth(getAdminApp()).verifyIdToken(idToken);
+}
+
+export async function upsertPageVisitFromServer(payload: {
```
romman8: This file starts with `'use server'`, so every export becomes a callable server action. This function takes `userId` from its caller and checks no auth itself, so it is an unauthenticated write path into `page_visits` — exactly what the rules block on the client side. `verifyFirebaseIdToken` is exposed the same way.

The route imports both directly and needs no action semantics, so please move them into a plain module with `import 'server-only'`.

web/lib/firebase/firebase-admin.ts
```diff
@@ -29,6 +30,67 @@ import type { ShareableChatSessionSnapshot, Tenant } from './firebase.types';
...
+      : 0;
+  const visibleMs = Math.max(existingVisibleMs ?? 0, payload.visibleMs);
+
+  const data: Record<string, unknown> = {
+    user_id: payload.userId,
+    last_seen_at: Timestamp.now(),
+    visible_ms: visibleMs,
+  };
+  if (!existing.exists) {
+    data.started_at = payload.startedAtMs
```
romman8: When a request has no `started_at_ms`, the doc is created with `Timestamp.now()`. The client later sends its own `started_at`, which then differs and fails the `started_at == resource.data.started_at` update rule. After that the client can never update this visit again.

Please require `started_at_ms` here, or stop sending `started_at` from the client once the doc may already exist.

firebase/firestore.rules
```diff
@@ -86,6 +86,29 @@ service cloud.firestore {
...
+        && request.resource.data.keys().hasAll([
+          'user_id',
+          'started_at',
+          'last_seen_at',
+          'visible_ms'
+        ])
+        && request.resource.data.user_id == request.auth.uid
+        && request.resource.data.visible_ms is number
+        && request.resource.data.visible_ms >= 0;
+      allow update: if request.auth != null
```
romman8, line 134: The update rule checks ownership and monotonic growth, but it does not limit values. A user can set `visible_ms` to any large number, backdate `last_seen_at`, write any `chat_session_ids`, and add any extra field up to the 1 MB doc limit.

These docs are study data, so please add an upper bound on `visible_ms` (the same 7 days the client clamps to) and an allowlist of writable fields.

web/lib/firebase/firebase.ts
```diff
@@ -71,6 +75,68 @@ export async function getAuthHeader(): Promise<Record<string, string>> {
...
+  if (payload.tenantId) {
+    data.tenant_id = payload.tenantId;
+  }
+  if (payload.chatSessionId) {
+    data.chat_session_ids = arrayUnion(payload.chatSessionId);
+  }
+  await setDoc(doc(db, 'page_visits', payload.visitId), data, { merge: true });
+}
+
+export async function attachChatSessionToPageVisit(
```
romman8, line 116: `getCurrentVisitId()` also falls back to `sessionStorage`, but this function adds the create fields only when a live runtime snapshot exists. In that window the write has no `started_at`/`visible_ms`, fails `hasAll` in the create rule, and the chat session keeps a `visit_id` that points to no doc.

It is also fire-and-forget and races the first flush, so both writes can try to create the same doc.

web/lib/page-visit/page-visit.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+function getStorage(): StorageLike | null {
+  if (storageOverride !== undefined) {
+    return storageOverride;
+  }
+  if (typeof sessionStorage === 'undefined') {
+    return null;
+  }
+  return sessionStorage;
```
romman8, line 57: The visit id lives in `sessionStorage`, so two open tabs of the same user create two visits. The description counts several visits per `user_id` as return visits, so multi-tab use will inflate that number.

If return visits are a study output, please define them by time gap, or store the relation between tabs.

firebase/firestore.rules
```diff
@@ -86,6 +86,29 @@ service cloud.firestore {
...
+        && resource.data.user_id == request.auth.uid
+        && request.resource.data.user_id == resource.data.user_id
+        && request.resource.data.visible_ms is number
+        && request.resource.data.visible_ms >= resource.data.visible_ms
+        && request.resource.data.started_at == resource.data.started_at;
+      allow delete: if false;
```
romman8, line 142: Visits are never deleted and there is no TTL policy, so this collection grows without limit. Please add a Firestore TTL policy on `last_seen_at` that matches the retention we promise in the privacy policy.

web/lib/page-visit/page-visit.ts
```diff
@@ -0,0 +1,200 @@
...
+  return clampVisibleMs(
+    runtime.snapshot.visibleMs + (now - runtime.segmentStartedAt),
+  );
+}
+
+export function getPageVisitSnapshot(): PageVisitSnapshot | null {
+  return runtime?.snapshot ?? null;
+}
+
+export function contextIdFromPath(pathname: string): string | undefined {
```
romman8, line 213: Nit: this denylist covers all current top-level routes, but the next new page will silently produce a wrong `context_id`. Reading the `[contextId]` route param instead would not drift.

web/lib/page-visit/page-visit-flush.ts
```diff
@@ -0,0 +1,137 @@
...
+  embedded?: boolean;
+  idToken?: string;
+};
+
+let flushInFlight = false;
+let flushQueued = false;
+let queuedBeacon = false;
+let queuedContext: PageVisitFlushContext | null = null;
+
+export function resetPageVisitFlushForTests(): void {
```
romman8, line 24: Nit: `resetPageVisitFlushForTests` is never used. The flush queue is the only stateful part with no coverage — the coalescing and the queued beacon flag would be worth one test.

web/components/providers/page-visit-provider.tsx
romman8: Fixed. The server layout now passes `embedded={IS_EMBEDDED}` into the client provider, so the embed deployment actually records the flag.

web/lib/page-visit/page-visit-request.ts
romman8: Fixed. Tokens now go through `optionalIdToken` with a 4096-char cap. `POST /api/page-visit` returns 204 against the emulator.

web/lib/page-visit/page-visit-flush.ts
romman8: Fixed. Hide now stops the segment, checkpoints, then `sendBeacon` / keepalive first. The client Firestore write is best-effort after that.

web/lib/page-visit/page-visit.ts
romman8: Fixed. `checkpointVisibleMs()` persists the flushed value and resets the open segment start, so a crash restore cannot fall behind Firestore.

web/lib/firebase/firebase-admin.ts
romman8: Fixed. Admin init lives in `firebase-admin-app.ts` (`import 'server-only'`). Visit helpers are in `page-visit-admin.ts`. The route no longer imports from the `'use server'` module.

web/lib/firebase/firebase-admin.ts
romman8: Fixed. Server create now requires `started_at_ms`. No `Timestamp.now()` fallback.

firebase/firestore.rules
romman8: Fixed. Writes are `hasOnly`-allowlisted, `visible_ms` is capped at 7 days, and `chat_session_ids` is limited to 100.

web/lib/firebase/firebase.ts
romman8: Fixed. Attach loads the snapshot and writes the create fields when the doc is missing. If it already exists we only `arrayUnion` the session id, so we cannot decrease `visible_ms`.

web/lib/page-visit/page-visit.ts
romman8: Leaving this as-is. A visit is tab-scoped visible dwell on purpose. Return visits should be a time gap on `user_id`, not a raw visit count.

firebase/firestore.rules
romman8: Leaving this as a follow-up. Firestore TTL is a console/ops setting, not a rules change, and we should not invent `expire_at` until the privacy-policy retention window is locked.

web/lib/page-visit/page-visit.ts
romman8: The provider now prefers `useParams().contextId` when present. It lives in the root layout, so Next will not see the child `[contextId]` param there — `contextIdFromPath` stays as the fallback.

web/lib/page-visit/page-visit-flush.ts
romman8: Added a coalescing test. The queued-beacon flag is gone because unload now beacons synchronously first.

RobinFrasch (approved):
tested by human and by AI, lg

PR #178, opened 2026-09-13
Make chat retrieval async so Qdrant queries do not block the event loop

### LINKED ISSUE #173: Make sure that there are no blocking synchronous requests (e.g. for retrieval queries) in the ai chat service API

romman8:
## Summary
- Switch chat `retrieve()` / `retrieve_two_pass()` to `AsyncQdrantClient` and `aembed_query`, and await them from the chat stream and voting-behavior route instead of `asyncio.to_thread` (or blocking on the loop).
- Offload leftover sync `query_points` / `get_collections` calls in the unused V1 `vector_store_helper` async helpers onto the thread pool.
- Ingestion keeps the synchronous `QdrantClient`.

Fixes https://github.com/wahl-chat/wahl-chat-app/issues/173

## Other blocking I/O on the chat path
After this change, the request-path I/O we scanned is already async:
- embeddings: `embed.aembed_query`
- Firestore: `firestore_async` helpers
- LLMs: `ainvoke` / `astream`
- Perplexity: `AsyncOpenAI`

Startup `REQUIRE_CORPUS` still uses a sync Qdrant count (once, at boot). Ingestion runners stay sync on purpose.

## Test plan
- [x] `pytest tests/ingestion/test_retrieve.py tests/test_retrieve.py tests/test_chat_service.py tests/test_source_filter.py tests/test_chat_sse.py`
- [x] ruff + mypy on the changed modules
- [ ] Smoke a live chat turn against local Qdrant (`make stores-up` + `make dev`) and confirm answers still stream with citations


ai-backend/src/ingestion/retrieve.py
```diff
@@ -10,7 +10,8 @@
 
 This module is STANDALONE — it does NOT import chat_service or vector_store_helper.
 The V1 retrieval path is untouched. Retrieval is testable without a live Gemini
-call (call retrieve() directly with a mocked embed).
+call (await retrieve() directly with a mocked embed). Chat callers await these
+functions on the request path; ingestion keeps the sync QdrantClient.
```
romman8: keep comments concise and timeless. Improve the overall comment to fulfill this requirement and drop the slop

ai-backend/src/ingestion/retrieve.py
```diff
@@ -91,28 +101,28 @@ def _get_embed() -> Embeddings:
...
     """Embed a query string into a dense vector using the locked embedding model.
 
-    Branches on the ``embed_query`` interface FIRST (covers OpenAIEmbeddings and any
-    LangChain embeddings — these are NOT callable), then falls back to treating
-    ``embed_fn`` as a plain callable (a bare function mock in tests). A
-    ``callable()``-first gate would make OpenAIEmbeddings (non-callable) skip the
-    embed_query path entirely, which is contradictory.
+    Prefers ``aembed_query`` so LangChain clients (OpenAIEmbeddings, Gemini) do
+    not block the event loop. Falls back to ``embed_query`` then to a plain
+    callable (bare function mocks in tests). A ``callable()``-first gate would
```
romman8: Is there any reason for this fallback to the synchronous "embed_query"? This should not happen within an async function. Drop the fallback if not necessary and cut the slop out of the comment. The comment must be concise and timeless and should NOT narrate what the code does

ai-backend/src/ingestion/retrieve.py
```diff
@@ -422,7 +432,8 @@ def retrieve(
                         ``point.score`` in the plain branch. Used by
                         ``retrieve_two_pass`` to apply recency decay on top of the
                         retrieve() ranking score.
-        _client:        Optional QdrantClient override (for tests).
+        _client:        Optional Qdrant client override (for tests). Accepts
+                        AsyncQdrantClient, sync QdrantClient, or a mock.
```
romman8: Would prefer splitting into a synchronous and an asynchronous function rather than mixing into one so that it is always clear for the caller what to expect and avoid any surprising blocking calls in asynchronous code

ai-backend/src/ingestion/retrieve.py
romman8: Tightened the module docstring: dropped the V1/ingestion asides and kept the standalone + testability note.

ai-backend/src/ingestion/retrieve.py
romman8: Dropped the sync `embed_query` fallback. The remaining callable branch is only for injected test doubles (`_embed_fn`), which are local functions, not I/O.

ai-backend/src/ingestion/retrieve.py
romman8: `retrieve()` is async-only now: it always awaits `AsyncQdrantClient.query_points`. There is no sync retrieve on the request path (ingestion never used this module), so a second sync function would be unused. Tests inject `AsyncMock` or a real async client.

PR #186, opened 2026-09-15
PledgeTracker study: consent, cohorts, gating, telemetry, questionnaire

miquelcamprubi:
Implements the Cambridge study (Prof. Vlachos / Prof. Stafford / Robin) around
PledgeTracker: does the feature change users' willingness to engage in
political debate? Runs **only** in `abgeordnetenhauswahl-berlin-2026` and
`landtagswahl-mecklenburg-vorpommern-2026`.

## What it does

- **Two-stage consent** (reviewed German copy, verbatim) before the first
  message in study contexts. „Nein" is remembered per user — never asked
  again, and **dismissing the dialog (ESC/backdrop/swipe) counts as a
  decline**, so everyone who was asked leaves a record. Both answers store
  `context_id` + `party_ids` (the parties selected when asked).
- **Deterministic 50/50 cohorts** at consent-accept: FNV-1a
  `hash(uid + salt) % 2`, persisted to `study_participants/{uid}` as the
  single source of truth for analysis. Same uid ⇒ same group, always.
- **Feature gate**: in study contexts, PledgeTracker renders only for
  consented *experimental* participants. Control **and non-consented users**
  see nothing (pre-exposure would contaminate later assignment). Outside the
  two study contexts nothing changes.
- **Kill switch**: Firestore `system_status/pledge_study { enabled }`.
  Missing doc, read error, or `false` ⇒ study off and PledgeTracker behaves
  exactly as on `develop`. Console-editable, no deploy needed. Can be swapped
  for Vercel Flags when that tooling lands — the gate takes a boolean.
- **Consent-gated telemetry** on the participant doc (append-only event log:
  `first_message`, `first_answer_completed`, `pledge_shown`,
  `pledge_modal_open/close`, `prompt_shown {trigger}`, `prompt_dismissed`,
  `questionnaire_clicked`). Chat sessions get stamped `study_group` +
  `is_pledge_study` so Firestore session metrics join to cohorts. Non-participants:
  strict no-op.
- **Questionnaire prompts, control-symmetric**: one 15 s idle timer after the
  first answer (identical for both cohorts — treatment can only shift timing
  through the user's own reading, never through group-dependent rules), a
  retry when the pledge modal closes, a 90 s long-stop for modal abandoners,
  hard cap of 2 prompts, none after a click. Opens **in-app** via
  `FilloutPopupEmbed` (form `s9muKGX2zcus`), carrying `user_id` +
  `chat_session_id` — so a response joins to the chat session and, through
  `page_visits.chat_session_ids`, to that visit's dwell time. The cohort is
  never passed, so participants can't unblind themselves.
- Firestore rule for `study_participants` (own-doc read/write) and an
  `AGENTS.md` runbook (ops, event schema, analysis joins).

## Tested

- End-to-end locally against the emulator: consent → cohort assignment →
  gating (both cohorts) → full telemetry chain verified doc-by-doc (timer
  fired at exactly +15 s; prompts stop after questionnaire click; non-study
  contexts unstamped) → Google Form opens with the uid prefilled.
- `tsc --noEmit` and `next lint`/`biome` clean; backend 967 passed.
  Note: 5 `test_seed_revalidate.py` failures are **pre-existing at
  `a05d53f`** (they pass in isolation — test-ordering/env pollution, not
  from this branch's commits).

## Remaining for production

1. Create prod `system_status/pledge_study { enabled: false }` — flipping it
   to `true` is the study launch, no deploy involved.
2. `firebase deploy --only firestore:rules`.
3. Pledge ingestion into prod (independent track): Cambridge's API is
   currently ~10–12 min/pledge with occasional 502s, so the full 252-pledge
   registry needs either lead time, more queue throughput, or a results dump
   from Cambridge.

## Added since first review

- Questionnaire is a **Fillout popup**, opened in-app like every other Fillout
  form here, with `user_id` + `chat_session_id`. The form id is **committed**,
  so a fresh clone and both deployments work with no env setup;
  `NEXT_PUBLIC_STUDY_QUESTIONNAIRE_URL` only overrides it for a test form. The
  kill switch remains the off-switch.
- Dismissing consent now counts as a decline; both answers record
  `context_id` + `party_ids`, so refusal can be modelled per election and per
  party rather than only counted.
- Consented participants are shown **no other survey** while the study is on
  (`survey-banner` self-suppresses). Cohort-independent, so both arms stay
  identical, and tied to the kill switch, so it returns for everyone afterwards.
- **Dev-only cohort toggle** (`ChatStudyDevBar`): local dev only, localStorage,
  never writes `study_participants`.
- First tests for `lib/pledge-study`.
- Separate commit: the emulator client SDK now uses `localhost` instead of a
  hardcoded `127.0.0.1`.

## Testing locally

```bash
cp web/.env.example web/.env          # emulator defaults work as-is
make stores-up
FIRESTORE_EMULATOR_HOST=localhost:8081 make seed-local
# turn the study on (the emulator UI is disabled, so write it over REST):
curl -X PATCH "http://localhost:8081/v1/projects/demo-wahl-chat/databases/(default)/documents/system_status/pledge_study" \
  -H 'Authorization: Bearer owner' -H 'Content-Type: application/json' \
  -d '{"fields":{"enabled":{"booleanValue":true}}}'
cd web && bun run dev
```

Open `/abgeordnetenhauswahl-berlin-2026/session`. Fully testable this way:
consent (accept / decline / dismiss) and the resulting `study_participants`
doc, the dev bar and cohort toggle, survey suppression, and the 15 s
questionnaire prompt with the Fillout popup carrying both parameters.

## Only testable against dev Qdrant + Firestore

The **PledgeTracker card itself**. It needs `source_type="pledge_record"`
points in Qdrant *and* the `pledges` collection in Firestore; a fresh local
stack has neither, so the cohort toggle has nothing visible to switch between.
Same for the `pledge_shown` / `pledge_modal_open|close` telemetry and the
`modal_close` / `longstop` questionnaire triggers.

Either ingest locally
(`FIRESTORE_EMULATOR_HOST=localhost:8081 make run-pledgetracker`) or point at
dev. ⚠️ **Pointing at dev Firestore means test consent writes land in the real
`study_participants` collection** — clear them before launch, or the cohort
data is contaminated.


web/components/chat/chat-pledge-tracker-button.tsx
```diff
@@ -101,7 +103,10 @@ function ChatPledgeTrackerButton({
       message_id: message.id,
       pledges,
     });
-  }, [message.id, partyId, pledges]);
+    // Study exposure log (no-op for non-participants): the researchers need
+    // per-participant "treated vs merely assigned" — this is that signal.
+    void recordStudyEvent('pledge_shown');
```
romman8, line 108: Why do we need this on top of `track('pledge_tracker_impression'`?
We already know who is part of the experiment but outside of the PledgTracker Variant and can simply check if they sent a message within the respective election in the relevant time frame, no?

web/components/chat/chat-study-consent.tsx
```diff
@@ -0,0 +1,180 @@
...
+              </p>
+            </div>
+            <ResponsiveDialogFooter>
+              <div className="flex w-full flex-col gap-2 sm:flex-row">
+                <Button variant="outline" className="w-full" onClick={decline}>
+                  Nein
+                </Button>
+                <Button
+                  className="w-full"
+                  onClick={() => setStep('einverstaendnis')}
```
romman8: nit: let's avoid German like "einverstaendnis" in states within the code

web/components/chat/chat-study-questionnaire-prompt.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * Questionnaire prompt engine — the ratified trigger design, identical for
+ * BOTH cohorts (control symmetry; "pledge card seen" is deliberately not an
+ * arming condition):
+ *
+ * - timer:       QUESTIONNAIRE_DELAY_MS after the FIRST answer completes,
+ *                fired only while idle (no streaming, no open pledge modal —
+ *                the effect re-arms whenever idleness returns). First prompt
+ *                only.
```
romman8: "First prompt only."
Wouldn't this mean that the questionnaire is never shown if the model quickly asks a follow-up question?
Could be better to start this timer after receiving the response to the first message and only interrupting it in case the user opens the pedgetracker. Wdyt?

romman8: Or is it only implemented like that and "First prompt only" means that the timer is already started after the first message of a user?

web/components/chat/chat-study-questionnaire-prompt.tsx
```diff
@@ -0,0 +1,214 @@
...
+ *                right after the treatment episode ends, never inside it.
+ *                Also the single retry path after a dismissed prompt.
+ * - longstop:    a pledge modal open for MODAL_LONGSTOP_MS prompts anyway
+ *                (catches abandoned tabs mid-modal).
+ *
+ * Hard cap MAX_PROMPTS across reloads (count is derived from the persisted
+ * event log on hydrate). Timing asymmetry between cohorts can only arise
+ * from the participant's own use of the feature — part of the treatment —
+ * never from group-dependent scheduling rules.
+ */
```
romman8, line 46: Can we rewrite this slop comment in Simplified Technical English?
My brain is not processing sentences like "Timing asymmetry between cohorts can only arise
 * from the participant's own use of the feature — part of the treatment —
 * never from group-dependent scheduling rules" after a day of work

web/components/chat/chat-study-questionnaire-prompt.tsx
```diff
@@ -0,0 +1,214 @@
...
+              Deine Rückmeldung zählt
+            </ResponsiveDialogTitle>
+            <ResponsiveDialogDescription>
+              Kurzer Fragebogen zur Studie
+            </ResponsiveDialogDescription>
+          </ResponsiveDialogHeader>
+          <div className="px-4 text-sm md:px-0">
+            <p>
+              Danke, dass du bei unserer Studie mitmachst! Wir würden dir gern
+              ein paar kurze Fragen zu deinem heutigen Besuch stellen — es
```
romman8: I am a bit allergic to em-dashes these days. Might be more trustworthy if we use wording without

web/components/chat/survey-banner.tsx
```diff
@@ -78,6 +80,18 @@ function SurveyBanner() {
...
+  // study questionnaire. Two surveys competing for the same goodwill depress
+  // the response rate on the one the study depends on.
+  //
+  // The predicate deliberately ignores the cohort, so control and experimental
+  // are exposed identically (differential survey exposure would confound the
+  // experiment), and it is tied to the kill switch rather than to consent
+  // alone, so the banner returns for everyone the moment the study is switched
+  // off instead of staying muted forever for past participants. Users who
+  // declined keep the normal product throughout.
+  if (studyEnabled && studyConsent === 'accepted') return null;
```
romman8: When is the survey banner shown right now? I would actually be in favor of simply disabling it completely for the time of the study to avoid any competition with the consent popup and the study questionnaire

web/lib/firebase/firebase.types.ts
```diff
@@ -91,6 +95,40 @@ export type LlmSystemStatus = {
...
+
+/**
+ * study_participants/{uid} — consent-gated per-participant record for the
+ * PledgeTracker study. Counts and firsts are DERIVED from `events` at
+ * analysis time (min/count per type), so the doc stays append-mostly.
+ */
+export type StudyParticipant = {
+  consent_answer: 'accepted' | 'declined';
+  consent_at: Timestamp;
+  group?: 'control' | 'experimental';
```
romman8: nit: we keep repeating the strings 'control' | 'experimental'. Might makes sense to have a type for this

web/lib/pledge-study/dev-cohort-override.ts
```diff
@@ -0,0 +1,54 @@
+import type { StudyCohort } from './study-config';
+
+/**
+ * Dev-only cohort override for the PledgeTracker study.
+ *
+ * The cohort is a deterministic hash of the anonymous uid, so without this the
+ * only way to see the other arm locally is to clear site data until the hash
+ * lands the other way. The override is kept in localStorage (so it survives a
```
romman8: hmm, my preferred solution would be to define a query param that allows us to set the desired variant (declined participation, control or PledgeTracker variant) so that we can also easily test the variant we desire in dev and prod environments

web/components/chat/chat-pledge-tracker-button.tsx
RobinFrasch: Might make it more comfortable to analyse the data later if we see directly if they received a pledge or asked something where no pledge was available. So mainly convenience.

web/components/chat/chat-study-consent.tsx
RobinFrasch: Agree, lets use English phrases.

web/components/chat/chat-study-questionnaire-prompt.tsx
RobinFrasch: Starting the timer after the first response is complete makes more sense, since that is realistically when the reading time starts. Also any lag or long generation time won't be an issue then.

web/components/chat/chat-study-questionnaire-prompt.tsx
RobinFrasch: Agree, lets simplify this.

web/components/chat/chat-study-questionnaire-prompt.tsx
RobinFrasch: Done

web/components/chat/survey-banner.tsx
RobinFrasch: Agree, lets disable all survey banners except for the one related to the pledge tracker study

web/lib/firebase/firebase.types.ts
RobinFrasch: Agree, generate a type for the groups. Be sure to differentiate between users who agree to participate in the study or not (experimental or regular) and then whether they are in the control group or manipulation group. Be sure to not double load words

web/lib/pledge-study/dev-cohort-override.ts
RobinFrasch: only issue might be, that users could see in which group they are if they inspect the query URL. But most people won't so its fine. Let's remove the toggle function I implemented for dev and implement this handling via the query param. Be sure to maintain the 50:50 split between control or manipulation cohort.

web/components/chat/chat-pledge-tracker-button.tsx
RobinFrasch: Kept as is. The two events answer different questions. track('pledge_tracker_impression') is aggregate analytics. pledge_shown is per participant, so in the analysis we can tell apart a user who was shown a pledge from a user who asked something where no pledge existed. Deriving that later from message data would be extra work. So this is mainly for convenience when we analyse the study.

web/components/chat/chat-study-consent.tsx
RobinFrasch: Fixed. The state is now 'ask' | 'consent' instead of 'ask' | 'einverstaendnis'. German is now only used in the text that participants read.

web/components/chat/chat-study-questionnaire-prompt.tsx
RobinFrasch: Two parts here.

First, the timer already started after the first answer was complete, not after the first user message. So that part was the comment being unclear, not a bug. The comment is now rewritten.

Second, your other point was correct and is fixed. The timer used to pause while a follow-up answer was still streaming. It no longer does. Only an open PledgeTracker modal delays it now. So the questionnaire appears 15 seconds after the first answer completes, even if the user has already asked something else.

One note on wording: "prompt" here means the dialog that asks the user to open the questionnaire. It is not a model prompt.

web/components/chat/chat-study-questionnaire-prompt.tsx
RobinFrasch: Rewritten in plain English

web/components/chat/survey-banner.tsx
RobinFrasch: Done, and wider than the original suggestion. While the study is on, no other survey is shown anywhere in the app:

- the general feedback banner in the chat
- the newsletter question after login
- the Wahl-Swiper feedback card
- 
A new StudyStatusProvider holds one subscription to the kill switch, and useStudyRunning() is used at each of those places. The rule ignores cohort and consent, so no group is treated differently. Everything comes back as soon as the kill switch is turned off.

Two things were left on purpose: the Wahl-Swiper "experimental" disclaimer (it is a warning, not a survey) and the rate limit notice (it is a functional message).

One cost to be aware of: every visitor now opens one Firestore listener on system_status/pledge_study, not only study participants.

web/lib/firebase/firebase.types.ts
RobinFrasch: Done. There are now two separate types in web/lib/pledge-study/types.ts:

StudyParticipation = 'experimental' | 'regular' — does the user take part in the study at all?
StudyCohort = 'control' | 'manipulation' — which group is a participant in?
So "experimental" no longer means "sees the feature". The group that sees PledgeTracker is now called manipulation.

The Firestore fields changed with it: group is now cohort, study_group on chat sessions is now study_cohort, and participation is new. No migration was needed because study_participants was empty.

web/lib/pledge-study/dev-cohort-override.ts
RobinFrasch: Done. The dev bar and the localStorage override are deleted. There is now a URL parameter that works in dev and in production:

?sg=a consented, control group
?sg=b consented, manipulation group
?sg=x declined participation
?sg=off clears it
The values cannot be read as a group name, so a participant does not see their group in the URL. There is no signing or validation, because the repo is public and anything stronger would not actually help.

An override also switches the study on for that browser only. Without this it could not be tested in production before launch, because the kill switch is off there.

Forced sessions are marked in Firestore with assignment_source: 'override' and must be excluded from the analysis. The normal 50:50 hash is unchanged.

Two things to know: use a fresh browser profile or incognito for these links, because opening one as a real participant marks that participant as excluded. And an override link ignores the kill switch.

romman8 (approved):

